PR #13, opened 2019-06-12
Deploy auth actions from CLI

Himavanth:
Deploy auth actions from CLI through a questionaire

## Types of changes

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "author": "Adobe Inc.",
   "bugs": "https://github.com/adobe/aio-cli-plugin-cna/issues",
   "dependencies": {
+    "@adobe/aio-cli-config": "^1.0.7",
     "@adobe/io-cna-scripts": "^0.0.5",
     "@oclif/command": "^1.5.11",
     "@oclif/config": "^1.12.9",
@@ -60,6 +61,7 @@
   "scripts": {
     "postpack": "rm -f oclif.manifest.json",
     "posttest": "eslint .",
+    "eslint-fix": "eslint . --fix",
     "prepack": "oclif-dev manifest && oclif-dev readme",
     "test": "jest -c ./jest.config.js",
     "version": "oclif-dev readme && git add README.md"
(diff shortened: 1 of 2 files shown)


src/commands/cna/create-auth.js
```diff
@@ -0,0 +1,272 @@
...
+    })
+    await configureAuth(destDir, IMS_AUTH_TYPE, IMS_AUTH_PERSIST)
+    await IMS_AUTH_TYPE === 'code' ? deployAuth(destDir) : deployJWTAuth(destDir)
+  }
+}
+
+async function cloneRepo () {
+  return new Promise(function (resolve, reject) {
+    cli.action.start('cloning auth library')
+    const child = spawn('git', ['clone', 'https://github.com/adobe/adobeio-cna-lib-auth-ims'])
```
purplecabbage: This means we could be creating a git repo inside the developers app, which might also be in a git repo.
I think ultimately we just want the action source files and some additional config added to the developers project, and they could then commit that to their own version control.

package.json
```diff
@@ -60,6 +61,7 @@
   "scripts": {
     "postpack": "rm -f oclif.manifest.json",
     "posttest": "eslint .",
+    "eslint-fix": "eslint . --fix",
```
purplecabbage: Why not just type `npx eslint . --fix` instead of `npm run eslint-fix`? It's even less keystrokes.

src/commands/cna/create-auth.js
```diff
@@ -0,0 +1,272 @@
...
+const path = require('path')
+const fs = require('fs-extra')
+const spawn = require('cross-spawn')
+const which = require('which')
+const config = require('@adobe/aio-cli-config')
+const inquirer = require('inquirer')
+
+class CNACreateAuth extends CNABaseCommand {
+  async run () {
+    this.parse(CNACreateAuth)
```
purplecabbage: nit: parse return values are not used, so why parse?

PR #24, opened 2019-07-09
[ACNA-349] Parse path for package name

Himavanth:
Parse path for package name

## Description

Parsing the path to pick a valid package name that can work for an s3 folder

## Related Issue

https://github.com/adobe/aio-cli-plugin-cna/issues/23

## Types of changes

## Checklist:

### DIFF
--- a/src/commands/cna/init.js
+++ b/src/commands/cna/init.js
@@ -66,7 +66,7 @@ class CNAInit extends CNABaseCommand {
         ]
       }])
     }
-    
+
     let packageName = destDir.substr(destDir.lastIndexOf('/') + 1)
     await this.copyBaseFiles(destDir, packageName, flags.yes)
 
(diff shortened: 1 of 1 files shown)


moritzraho:
Hi @Himavanth 
Just a few comments :) 
- Can you add some checks when the user is prompted for the package name:  https://github.com/adobe/aio-cli-plugin-cna/blob/2c423cf50c42ade3ed5e10d37694b3622569c3ff/src/commands/cna/init.js#L112, we could output a message saying that the package name was normalized if the users entered package name is invalid
- ideally we should make sure that the packagename is accepted by both s3 and I/O Runtime. to be safe we could do something like accepting only lowercase and numbers and '-' 
- Can you add a test to cover the change?

Himavanth:
Yes valid points. I see there are other validations pending for the init command and the codecov is only 13% right now. It will be a larger story so i didn't want to do that in a bug fix PR. Created a story for that. https://jira.corp.adobe.com/browse/ACNA-355

PR #59, opened 2019-11-04
add -y|-yes and -s flags

Yu1986:
## Description

add -y | -yes so will choose the default answers, skip prompting questions
add -s will skip the install 
## Types of changes

## Checklist:

### DIFF
--- a/src/commands/app/gen.js
+++ b/src/commands/app/gen.js
@@ -1,3 +1,4 @@
+/* eslint-disable camelcase */
 /*
 Copyright 2019 Adobe. All rights reserved.
 This file is licensed to you under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


src/commands/app/gen.js
```diff
@@ -17,11 +18,22 @@ const { flags } = require('@oclif/command')
...
 class Gen extends BaseCommand {
   async run () {
-    // const { args } = this.parse(Gen)
+    const { args, flags } = this.parse(Gen)
     debug('creating new app with gen command')
+    let skip_prompt = false
+    if (flags.yes) {
+      skip_prompt = true
+    }
+    let skip_install = false
```
purplecabbage: This is a new flag, not a requirement.

src/commands/app/gen.js
```diff
@@ -17,11 +18,22 @@ const { flags } = require('@oclif/command')
...
+    if (flags.skill_install) {
```
purplecabbage: skill should be skip, but ultimately this is un-needed code

src/generators/createGenerator.js
```diff
@@ -28,6 +35,11 @@ will contain static assets to be uploaded to cloud storage. If you
 have a build process use your build's output directory.
 What folder do you want to use as your public web assets directory?
 `
+    let showPrompt = true
+    if (this.options.skip_prompt) {
+      showPrompt = false
+    }
+
```
purplecabbage: `const showPrompt = !this.options.skip_prompt`

src/generators/createGenerator.js
```diff
@@ -115,6 +140,9 @@ What folder do you want to use as your public web assets directory?
   }
 
   async install () {
+    if (this.options.skip_install) {
+      return
+    }
```
purplecabbage: We should only ever skip-install if the user replies No to the prompt
If the --yes flag is present, we don't prompt at all, so just install ...

Yu1986: fixed

PR #170, opened 2020-03-06
Handle Undeploy for UI only apps

sandeep-paliwal:
AIO scripts will now throw error when undeploy actions is called but there is no backend.
Handle this gracefully in undeploy command and not call scripts undeploy unless we have a valid backend

## Types of changes

## Checklist:

- [x ] I have signed the [Adobe Open Source CLA](http://opensource.adobe.com/cla.html).
- [x ] My code follows the code style of this project.
- [ x] I have read the **CONTRIBUTING** document.
- [x ] I have added tests to cover my changes.


src/commands/app/undeploy.js
```diff
@@ -59,7 +60,11 @@ class Undeploy extends BaseCommand {
 
       // undeploy
       if (!flags.static) {
-        await scripts.undeployActions()
+        if (fs.existsSync('actions/')) {
```
moritzraho: @sandeep-paliwal can we change this to fs.existsSync('manifest.yml') so we are aligned with app-scripts and https://github.com/adobe/aio-cli-plugin-app/pull/171

PR #186, opened 2020-04-09
Support for Get url command

sandeep-paliwal:
Added command to return Action URLs

## Description

For all actions in manifest
Specific action using 'action' option
CDN URLs for actions using 'cdn' flag

## How Has This Been Tested?

unit tests
CLI run

## Types of changes

- [ x] New feature (non-breaking change which adds functionality)

## Checklist:

- [x ] I have signed the [Adobe Open Source CLA](http://opensource.adobe.com/cla.html).
- [ x] My code follows the code style of this project.
- [x ] I have read the **CONTRIBUTING** document.
- [x ] I have added tests to cover my changes.
- [x ] All new and existing tests passed.


src/commands/app/get-url.js
```diff
@@ -0,0 +1,66 @@
...
+      if (urls.runtime) {
+        this.log('Runtime URLs')
+        Object.entries(urls.runtime).forEach(([key, value]) => {
+          this.log(chalk.blue(chalk.bold(`${key} `)) + ' - ' + chalk.blue(chalk.bold(`${value} `)))
+        })
+      }
+
+      if (urls.cdn) {
+        this.log('CDN URLs')
+        Object.entries(urls.cdn).forEach(([key, value]) => {
```
moritzraho, line 42: Question: should we make the log JSON parseable ?

sandeep-paliwal: is thing something we do for other commands o/o in general? If yes, then I can modify the PR. Else we can have a separate story to address all command o/p. WDYT?

moritzraho: yes ok let's do this as part of another PR.
maybe we could have something like a global --json flag in the future

sandeep-paliwal: looks like we have agreement. Resolving the conversation

src/commands/app/get-url.js
```diff
@@ -0,0 +1,66 @@
...
+        })
+      }
+      return urls
+    } catch (error) {
+      this.error(wrapError(error))
+    }
+  }
+}
+
+GetUrlCommand.description = 'Get action URL'
```
moritzraho, line 53: typo: Get action URLs

src/commands/app/get-url.js
```diff
@@ -0,0 +1,66 @@
...
+
+GetUrlCommand.flags = {
+  ...BaseCommand.flags,
+  cdn: flags.boolean({
+    description: 'Display CDN based action URLs'
+  })
+}
+
+GetUrlCommand.args = [
```
moritzraho, line 62: Here let's do as in the deploy command and use a multi flag like in the deploy command:
`aio app deploy -a action1 -a action2`
see https://github.com/adobe/aio-cli-plugin-app/blob/master/src/commands/app/deploy.js#L171

test/commands/app/geturl.test.js
```diff
@@ -0,0 +1,140 @@
...
+    const retVal = {
+      runtime: {
+        action: 'https://fake_ns.adobeioruntime.net/api/v1/web/sample-app-1.0.0/action'
+      }
+    }
+    mockScripts.getUrls.mockResolvedValue(retVal)
+    const urls = await command.run()
+    expect(command.error).toHaveBeenCalledTimes(0)
+    expect(mockScripts.getUrls).toBeCalledWith({})
+    expect(urls).toBe(retVal)
```
moritzraho, line 60: I think in all test cases we want to also test that the url is logged, as this is the intent of the command.
something like
`expect(command.log).toHaveBeenCalledWith(expect.stringContaining(url))`

src/commands/app/get-url.js
moritzraho: resolving as this is not in the requirements

src/commands/app/get-url.js
sandeep-paliwal: done

test/commands/app/geturl.test.js
sandeep-paliwal: done

PR #211, opened 2020-05-14
Use generator-aio-console in `app init`

shazron:
## Description

Addition of the console generator in app init. This requires user interaction in selecting the user's org, then project, then workspace -- to download the project config. The project config will then be imported and used.

To skip this user interaction, use the `--yes` or `--import` flag.

## How Has This Been Tested?

npm test
aio app init (should go through gen-console)
aio app init -y (skips gen-console)
aio app init --import config_file (skips gen_console)

## Types of changes

## Checklist:

### DIFF
--- a/src/commands/app/init.js
+++ b/src/commands/app/init.js
@@ -16,6 +16,8 @@ const fs = require('fs-extra')
 const aioLogger = require('@adobe/aio-lib-core-logging')('@adobe/aio-cli-plugin-app:init', { provider: 'debug' })
 const { flags } = require('@oclif/command')
 const { validateConfig, importConfigJson, loadConfigFile, writeAio } = require('../../lib/import')
+const { getToken } = require('@adobe/aio-lib-ims')
+const { CLI } = require('@adobe/aio-lib-ims/src/context')
 
 class InitCommand extends BaseCommand {
   async run () {
@@ -44,12 +46,25 @@ class InitCommand extends BaseCommand {
     }
 
     const env = yeoman.createEnv()
+    let res
 
     this.log(`You are about to initialize the project '${projectName}'`)
 
+    if (!flags.import) {
+      const accessToken = await getToken(CLI)
+
+      env.register(require.resolve('@adobe/generator-aio-console'), 'gen-console')
+      res = await env.run('gen-console', {
+        'access-token': accessToken
+      })
+
+      // trigger import
+      flags.import = 'console.json'
+    }
+
     // call code generator
     env.register(require.resolve('@adobe/generator-aio-app'), 'gen')
-    const res = await env.run('gen', {
+    res = await env.run('gen', {
       'skip-install': flags['skip-install'],
       'skip-prompt': flags.yes,
       'project-name': projectName,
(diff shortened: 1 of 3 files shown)


src/commands/app/init.js
```diff
@@ -28,9 +34,30 @@ class InitCommand extends BaseCommand {
...
+        env.register(require.resolve('@adobe/generator-aio-console'), 'gen-console')
+        res = await env.run('gen-console', {
+          'destination-file': generatedFile,
+          'access-token': accessToken,
+          'ims-env': imsEnv
+        })
+        // trigger import
+        flags.import = generatedFile
+      } catch (e) {
+        console.log(chalk.red(e.message))
```
purplecabbage, line 56: not this.error() ? we don't exit, just continue?

shazron: yup, as specified by @meryllblanchet

src/commands/app/init.js
```diff
@@ -28,9 +34,30 @@ class InitCommand extends BaseCommand {
 
     aioLogger.debug('creating new app with init command ', flags)
 
+    // default project name and services
     let projectName = path.basename(process.cwd())
     let services = 'AdobeTargetSDK,AdobeAnalyticsSDK,CampaignSDK,McDataServicesSdk,AudienceManagerCustomerSDK' // todo fetch those from console when no --import
 
+    if (!(flags.import || flags.yes)) {
+      const accessToken = await getToken(CLI)
+      const { env: imsEnv = 'prod' } = await context.getCli()
```
moritzraho, line 43: will this work if the CLI config is undefined ? (first time users)

src/commands/app/init.js
```diff
@@ -28,9 +34,30 @@ class InitCommand extends BaseCommand {
 
     aioLogger.debug('creating new app with init command ', flags)
 
+    // default project name and services
     let projectName = path.basename(process.cwd())
     let services = 'AdobeTargetSDK,AdobeAnalyticsSDK,CampaignSDK,McDataServicesSdk,AudienceManagerCustomerSDK' // todo fetch those from console when no --import
```
moritzraho, line 39: We don't need the hardcoded list anymore right? if yes at least let's remove the todo comment :)

shazron: the services is needed to bootstrap the `gen-app` generator, no? It's one of the options.

shazron: ... when `--import` is not used (or no console generator)

src/commands/app/init.js
shazron: oops, yes you are right it will fail here. I'll add a failing test and fix it.

src/commands/app/init.js
shazron: > 2 inline comments + the following:
> 
> * should we disallow --yes when no import ? cause then we would create a project without talking to console right? also in that case we won't need this logic https://github.com/adobe/aio-cli-plugin-app/blob/console-generator/src/commands/app/init.js#L93

if the console flow fails somehow for the dev, we still want them to be able to build a local app and try it out. 

I'm not familiar with the inner workings of the app plugin -- is `services` config key required somewhere?

moritzraho: > if the console flow fails somehow for the dev, we still want them to be able to build a local app and try it out.

ok yes this makes a lot of sense, I thought initially we would only build an app based on the console config, but providing a fallback option is good.

> I'm not familiar with the inner workings of the app plugin -- is services config key required somewhere?

Yes let's keep it in that case, it used by the `app add action` command actually: https://github.com/adobe/aio-cli-plugin-app/blob/master/src/commands/app/add/action.js#L25. (Thought that if we only allow apps based on a console config we would then use the config services  key provided by console.)

src/commands/app/init.js
shazron: updated with a fix commit

moritzraho:
lgtm

PR #233, opened 2020-05-18
Continue App init when Login timeouts

### LINKED ISSUE #225: Generator only allows logged in user
### Expected Behaviour

```
aio app init
```
should prompt me to log in. but if I don't have an Adobe ID, I should still be able to continue with this flow. 

### Actual Behaviour

Generator does not allow me to not log in. Times out if the user is not logged in. 

```
aio app init
Visit this url to log in: 
https://aio-login.adobeioruntime.net/api/v1/web/default/applogin?id=89125e9f&port=52227
Error: Timed out after 120 seconds.
    at Timeout._onTimeout (~/.local/share/@adobe/aio-cli/node_modules/@adobe/aio-lib-ims-oauth/src/login.js:51:14)
```

sandeep-paliwal:
Continue App init when Login timeouts.
Skip console generator in this case.
Fixes #225 

## How Has This Been Tested?
manual run and unit tests
## Types of changes

## Checklist:


src/commands/app/init.js
```diff
@@ -38,20 +38,30 @@ class InitCommand extends BaseCommand {
...
-        const generatedFile = 'console.json'
-        env.register(require.resolve('@adobe/generator-aio-console'), 'gen-console')
-        res = await env.run('gen-console', {
-          'destination-file': generatedFile,
-          'access-token': accessToken,
-          'ims-env': imsEnv
-        })
-        // trigger import
-        flags.import = generatedFile
+        const { accessToken, env: imsEnv } = await getCliInfo()
```
shazron, line 44: I'm confused why you just didn't take this line and just put in the bottom try catch, without resulting in this extra code. Having a custom error message is irrelevant.

sandeep-paliwal: OK. I thought in this case it would be good to log custom error. Anyway I can move this to the other try catch block

PR #279, opened 2020-08-06
[WIP] Handling subprocesses for app hooks

Himavanth:
### Current behaviour of app hooks 
- subprocess launched with npm run-script
- subprocess inherits same io streams as parent process
- SIGINT shuts down the parent process too
- > so any subprocess that uses SIGINT for exit cannot do so without closing the parent process down
- > this also introduces ambiguity at the stream level resulting in issues like the tty display bug
	

### Solution
- Let the subprocess exit without closing the parent process
- > we could remove the SIGINT listeners from parent process temporarily for this
- Use different IO streams for parent and child processes to remove any possibility of stream takeovers/closures

## Types of changes

## Checklist:

### DIFF
--- a/src/lib/app-helper.js
+++ b/src/lib/app-helper.js
@@ -51,8 +51,36 @@ async function runPackageScript (scriptName, dir, cmdArgs = []) {
   aioLogger.debug(`running npm run-script ${scriptName} in dir: ${dir}`)
   const pkg = await fs.readJSON(path.join(dir, 'package.json'))
   if (pkg && pkg.scripts && pkg.scripts[scriptName]) {
-    return execa('npm', ['run-script', scriptName].concat(cmdArgs), { cwd: dir, stdio: 'inherit' })
-  } else {
+
+    let termListeners = process.listeners('SIGINT')
+    process.removeAllListeners('SIGINT')
+    process.on('SIGINT', () => {
+      if(!childPro.killed) {
+        childPro.kill()
+      }
+    })
+
+    const childPro = execa('npm', ['run-script', scriptName].concat(cmdArgs), { cwd: dir, stdio: 'pipe' })
+    process.stdin.pipe(childPro.stdin)
+    childPro.stdout.pipe(process.stdout)
+    childPro.stderr.pipe(process.stderr)
+    
+    try {
+      await childPro
+    }catch(err){
+      // Ignore
+    }
+    process.stdin.unpipe()
+    process.stdin.resume()
+    // Adding the SIGINT listeners back
+    if(termListeners.length>0) {
+      process.removeAllListeners('SIGINT')
+      termListeners.forEach((listener) => {
+        process.addListener('SIGINT', listener)
+      })
+      termListeners = []
+    }
+} else {
     throw new Error(`${dir} does not contain a package.json or it does not contain a script named ${scriptName}`)
   }
 }
--- a/src/lib/app-helper.js
+++ b/src/lib/app-helper.js
@@ -61,6 +61,7 @@ async function runPackageScript (scriptName, dir, cmdArgs = []) {
     })
 
     const childPro = execa('npm', ['run-script', scriptName].concat(cmdArgs), { cwd: dir, stdio: 'pipe' })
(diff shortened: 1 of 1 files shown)


src/lib/app-helper.js
```diff
@@ -51,8 +51,37 @@ async function runPackageScript (scriptName, dir, cmdArgs = []) {
...
+    let termListeners = process.listeners('SIGINT')
+    process.removeAllListeners('SIGINT')
+    process.on('SIGINT', () => {
+      if(!childPro.killed) {
+        childPro.kill()
+      }
+    })
+
+    const childPro = execa('npm', ['run-script', scriptName].concat(cmdArgs), { cwd: dir, stdio: 'pipe' })
+    if(!childPro) return
```
moritzraho, line 64: missing braces

src/lib/app-helper.js
```diff
@@ -51,8 +51,37 @@ async function runPackageScript (scriptName, dir, cmdArgs = []) {
...
   const pkg = await fs.readJSON(path.join(dir, 'package.json'))
   if (pkg && pkg.scripts && pkg.scripts[scriptName]) {
-    return execa('npm', ['run-script', scriptName].concat(cmdArgs), { cwd: dir, stdio: 'inherit' })
-  } else {
+
+    let termListeners = process.listeners('SIGINT')
+    process.removeAllListeners('SIGINT')
+    process.on('SIGINT', () => {
+      if(!childPro.killed) {
+        childPro.kill()
```
moritzraho, line 59: shouldn't sigint also kill the main process ? seems like the user would need to CTRL +C two times

src/lib/app-helper.js
```diff
@@ -51,8 +51,37 @@ async function runPackageScript (scriptName, dir, cmdArgs = []) {
   aioLogger.debug(`running npm run-script ${scriptName} in dir: ${dir}`)
   const pkg = await fs.readJSON(path.join(dir, 'package.json'))
   if (pkg && pkg.scripts && pkg.scripts[scriptName]) {
-    return execa('npm', ['run-script', scriptName].concat(cmdArgs), { cwd: dir, stdio: 'inherit' })
-  } else {
+
+    let termListeners = process.listeners('SIGINT')
+    process.removeAllListeners('SIGINT')
```
shazron, line 56: why the side-effect of removing all listeners for the event? it's possible some script/code/hook might be listening as well for some reason

src/lib/app-helper.js
```diff
@@ -51,8 +51,37 @@ async function runPackageScript (scriptName, dir, cmdArgs = []) {
...
+    try {
+      await childPro
+    }catch(err){
+      // Ignore
+    }
+    process.stdin.unpipe()
+    process.stdin.resume()
+    // Adding the SIGINT listeners back
+    if(termListeners.length>0) {
+      process.removeAllListeners('SIGINT')
```
shazron, line 78: similar to my comment above

moritzraho:
One potential solution (not decided) as discussed this morning with the team:

- have a flag or additional hooks (e.g. `post-app-run-detached`) to treat long running processes as detached and cleanup on sigint (like the openwhisk jar in `--local`) instead of the main process waiting for the child process to terminate (which would be the case for short running child processes) 

we probably need to expose a distinction between short and long running commands in hooks:
- short running hooks might resolve a dependency on the run command (e.g. moving some files pre-built) and hence the main process needs to wait for the process to terminate
- long running hooks like the `asset-compute` command that spawns the NUI devtool UI need to run detached and be cleaned up on termination

jdelbick:
@moritzraho regarding your comment:

> long running hooks like the asset-compute command that spawns the NUI devtool UI need to run detached and be cleaned up on termination. Those could be started and terminated in the same way as the openwhisk jar subprocess in --local

what changes would we need to make in the devtool to support this?

moritzraho:
@jdelbick that's all hypothetical, but let's say we expose an additional `<pre/post>-app-run-detached` hook then the only thing you would need to do is use that hook to run the long running process instead of the current `<pre/post>-app-run`

Issue #283, opened 2020-08-31
Build & Deploy should be available as distinct CLI commands


meryllblanchet:
**Is your feature request related to a problem? Please describe.**
As of today, `aio app deploy` builds and deploys the whole app.
`aio app deploy --skip-build` only deploys the app.
`aio app deploy --skip-deploy` only builds the app.

This is confusing for users and we should provide a more obvious way to handle build and deploy separately. This can be very useful for CI/CD and other automation use-cases.

**Describe the solution you'd like**
Proposal:

`aio app build` builds the whole app
`aio app build -a|--action=action` builds an action
`aio app build --all-actions` builds all actions
`aio app build -s|--static` builds the SPA

`aio app deploy` behaves as of today if there's no build.
`aio app deploy` just deploys if there's a build.
`aio app deploy -b|--build` forces the build before the deploy (do we even need more fine-grained based?)

aiojbot:
JIRA issue created: https://jira.corp.adobe.com/browse/ACNA-906

meryllblanchet:
@Himavanth @purplecabbage let's also consider the `aio app run` command and its flags.

Himavanth:
Need to also consider multi-package manifests. (especially the `--action` flag)

PR #292, opened 2020-09-14
Separation of build and deploy commands

Himavanth:
## Description
Fixes https://github.com/adobe/aio-cli-plugin-app/issues/283
**Breaking changes**
- `aio app deploy` will not re-build if a previous build exists
- Should we remove `aio app deploy --skip-deploy` ?
- Should we rename `--skip-xxx` to `--no-xxx` ?

**New cmds/flags**
- `aio app deploy --force-build`
- `aio app build`, `aio app build --skip-actions`, `aio app build --skip-static`

## Types of changes

## Checklist:


sandeep-paliwal:
@Himavanth we use the deploy command in CI-CD git actions. I think the flag to force build should be added to those once we release this change.

shazron:
I'm resolving the conflicts and testing.

src/commands/app/deploy.js
```diff
@@ -171,9 +152,13 @@ Deploy.flags = {
   'skip-actions': flags.boolean({
     description: 'Skip action build & deploy'
   }),
+  'force-build': flags.boolean({
+    description: 'Forces a build even if one already exists',
+    exclusive: ['skip-build'],
+    default: false
```
shazron, line 158: should there be an "allowNo: true" property here?

Himavanth: The default is false here. So i think we don't need it. It is there in `aio app build` because default for `force-build` is true there. Also we are calling `build` from `deploy` with a `no-force-build` flag when `force-build` is not specified.

shazron (approved):

purplecabbage:
I would prefer to see both build+deploy use a BaseCommand implementation of build functionality.
BaseCommand is aware of flags, and this.log, and ...
My original complaint was simply that the flags were being passed into a library method to be processed, where the lib method should just be 'told' what to do.
imho, while it is 'cool!', commands calling each other is a bit of an anti-pattern, and something we should only use when there is no clearer way.

PR #301, opened 2020-09-29
ACNA-858 - support for app hook long-running processes via aio-run-detached

### LINKED ISSUE #270: aio app run messes up TTY echo on Ubuntu
### Expected Behaviour
Exiting AIO App run should restore TTY cleanly

### Actual Behaviour
Local echo is disabled and I can't see anything I'm typing.  I have to type `reset` to get my TTY console back to normal

### Reproduce Scenario (including but not limited to)
I am able to reproduce from Ubuntu within WSL2 on Windows 10.  I'm pretty sure it's an Ubuntu TTY thing because WSL2 is basically a full Ubuntu stack running in Windows hypervisor as a VM.

#### Steps to Reproduce
`aio app run` from a terminal shell (doesn't seem to happen for --local) and then after it is fully running, exit with CTRL-C

#### Environment Info
```
  System:
    OS: Linux 4.19 Ubuntu 19.10 (Eoan Ermine)
    CPU: (8) x64 Intel(R) Core(TM) i7-8650U CPU @ 1.90GHz
    Memory: 9.76 GB / 12.43 GB
    Shell: 5.0.3 - /bin/bash
  Binaries:
    Node: 14.5.0 - /usr/bin/node
    Yarn: Not Found
    npm: 6.14.6 - /usr/bin/npm
  Virtualization:
    Docker: 19.03.12 - /usr/bin/docker
  npmGlobalPackages:
    @adobe/aio-cli: 3.7.0
```

shazron:
closes #270 

## Description

Supports detached running of long-running processes (app hooks) via the [aio-run-detached](https://github.com/adobe/aio-run-detached)
 helper. Long running processes are cleaned up on process exit (command exit).

## How Has This Been Tested?

npm test
manual tests via app hooks with [aio-run-detached](https://github.com/adobe/aio-run-detached)

## Types of changes

## Checklist:


shazron:
Example long running process:
```
#!/usr/bin/env node

/** @private */
function sleep (ms) {
  return new Promise((resolve) => {
    setTimeout(resolve, ms)
  })
}

/** @private */
async function main () {
  for (var i = 0; i < 100; ++i) {
    await sleep(1000)
    console.log('i ->' + i)
  }
}

main().catch(e => {
  console.error(e)
  throw e
})
```

Usage, see [aio-run-detached](https://github.com/adobe/aio-run-detached/blob/master/README.md)

moritzraho:
the code lgtm but I did some manual testing and got the following behavior (not sure if it's expected, maybe I have a wrong setup):

- `"pre-app-run": "aio-run-detached ./script-long.js abc 123"`: doesn't log messages, in `./script-long.js` and if the script has an error while starting (e.g. no permissions set) it doesn't show it
- `"pre-app-run": "./script-long.js abc 123"`: shows logs this time but still not showing permission error 

while before `"pre-app-run": "./script-long.js abc 123"` was showing both logs and errors

shazron:
@moritzraho any errors are no-oped (the existing implementation has this https://github.com/adobe/aio-cli-plugin-app/blob/7e3183b128e4cbc7e194281083a3c8ecf4879894/src/commands/app/run.js#L54-L58)

long running processes are forked and detached, and requires detaching from the stdio of the process, so we won't see logs.

shazron:
found an issue. If the long running process has already terminated on its own, `aio` will try to kill it again, and this results in an exception. patching.

shazron:
Verified that if a script does not have execute permissions, it silently fails. It should log an error if the script does not have permissions. This is because of us silently eating the exception as described above.

The existing implementation calls `npm run-script` which catches this before it even reaches us. We execute the script directly now.

shazron:
@moritzraho Updated for review. Should report permissions errors now (need new @adobe/aio-run-detached v`1.1.2`)
See the updated `aio-run-detached` [README](https://github.com/adobe/aio-run-detached#logging) regarding logging, we can't use command redirect to files.

shazron:
@adobe/aio-run-detached@1.3.0 fixes the pathing issues, so this change can be merged now.

Issue #96, opened 2019-11-18
Add a log option to fetch logs indefinitely


cjelger:
It would be nice to have an option to fetch the logs "indefinitely", similar to what one can do with `tail -f` on some local log file.

So something like `aio app:logs -f` that would output all the logs indefinitely and would never return, except if you hit `ctrl + c`.

This would be very useful, because developers spend most of their time deploying code, checking the logs, deploying again, etc etc, so it would allow a developer to keep a "logs" terminal/console open in a separate window and get all the logs in almost real-time.

moritzraho:
Thanks @cjelger!

This somehow relates to https://github.com/adobe/aio-cli-plugin-app/issues/81 and there likely will be some common parts.

PR #303, opened 2020-10-05
New manifest and action level logs with new strip and tail options

Himavanth:
## Description
Fixes https://github.com/adobe/aio-cli-plugin-app/issues/96 among other enhancements as described in https://wiki.corp.adobe.com/display/AdobeCloudPlatform/Consolidation+of+logs+command+in+runtime+and+app+plugins

## Types of changes

## Checklist:


src/commands/app/logs.js
```diff
@@ -27,10 +29,25 @@ class Logs extends BaseCommand {
...
-      const config = this.getAppConfig()
+    let filterActions = []
+    if (flags.action) {
+      let actionName = flags.action
+      if (!actionName.includes('/')) {
+        actionName = config.ow.package + '/' + actionName
+      }
+      filterActions = [actionName]
+    } else {
+      Object.entries(config.manifest.full.packages).forEach((packageTuple) => {
```
purplecabbage, line 40: this is a little gnarly, not sure it could be any better though

src/commands/app/logs.js
```diff
@@ -46,11 +63,22 @@ Logs.flags = {
...
+    description: 'Fetch logs for a specific action',
+    default: '',
+    char: 'a'
+  }),
+  strip: flags.boolean({
+    char: 'r',
+    description: 'strip timestamp information and output first line only',
+    default: false
+  }),
+  tail: flags.boolean({
```
purplecabbage, line 77: is tail our final choice? should we alias it with `watch`?
what is the point of adding a default for -a? if they want to get logs for an action or actions, they would need to specify it ...

purplecabbage (approved):

src/commands/app/logs.js
Himavanth: thx @purplecabbage  . Removed action default and added aliases.

PR #314, opened 2020-10-28
add: config ui dir

rajarju:
## Local Development

[Sample FF project with CRA](https://github.com/rajarju/ff-react-app)

### 1. Single terminal window
- Have frontend code in a directory under ff app
- config .aio file to add
  - web - the path to the frontend directory
  - runScript - Package script to run frontend locally
- run `aio app run` to run the ui locally

### Issues with this approach
- The frontend script would use the same stdout as `aio app run`
- In case of `react-script` it would clear the logs before it runs, which would remove all the logs written by `aio app run` from stdout

### 2. Run frontend build as a separate process
- Have frontend code in a directory under ff app
- config .aio file to add
  - web - the path to the frontend directory
- run the custom frontend build/dev in a separate window (npm start)
- run `aio app run --skip-build` to run the ff app locally without frontend (parcel) build
- this would still create the config.json file inside `<cna.web>/src/config.json`

### Issues with this approach
- Devs will have to run frontend and backend (ff app) separately for local development

### 3. Running frontend script using aio-run-detached
- Have frontend code in a directory under ff app
- Install `aio-run-detached` as a dev dependency on frontend project
- add a custom dev script which runs the dev command using `aio-run-detached`
  - `"start:detached": "aio-run-detached react-scripts start"`

## Building and deploying

Ideally, developers should be able to configure their custom frontend src and build script 
and they should be able to run `aio app deploy` which would build and deploy the code with a single command

With the current version of app cli to deploy the frontend, developers have to
- run the build manually
- copy the build to ff build directory
- run `aio app deploy --skip-build`

### Todo
- Try using pre deploy hooks to run build

### DIFF
--- a/src/lib/app-helper.js
+++ b/src/lib/app-helper.js
@@ -18,6 +18,7 @@ const aioLogger = require('@adobe/aio-lib-core-logging')('@adobe/aio-cli-plugin-
 const { getToken, context } = require('@adobe/aio-lib-ims')
 const { CLI } = require('@adobe/aio-lib-ims/src/context')
 const fetch = require('node-fetch')
+const aioRunDetached = require('@adobe/aio-run-detached')
 
 /** @private */
 function isNpmInstalled () {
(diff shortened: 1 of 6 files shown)


src/lib/app-helper.js
```diff
@@ -81,6 +82,23 @@ async function runPackageScript (scriptName, dir, cmdArgs = []) {
...
+async function runPackageScriptDetached (scriptName, dir, cmdArgs = []) {
+  if (!dir) {
+    dir = process.cwd()
+  }
+  aioLogger.debug(`running npm run-script ${scriptName} in dir: ${dir}`)
+  const pkg = await fs.readJSON(path.join(dir, 'package.json'))
+  if (pkg && pkg.scripts && pkg.scripts[scriptName]) {
+    const command = pkg.scripts[scriptName].split(' ')
+
+    const child = aioRunDetached.run(command, dir)
```
rajarju, line 95: This is my attempt at running a detached process from aio app run
1. aioRunDetached needs to take a dir to cwd into
2. aioRunDetached.run should return the process so that it can be killed on exit

purplecabbage: This has since been changed and I think this pr needs to be updated or closed.

purplecabbage: Ping!! @rajarju ^^

rajarju:
- the changes are not valid anymore
- frontend dir is configurable now
- adding custom ui and build has been documented

PR #318, opened 2020-11-05
ACNA-942 - Windows: docker not found when running `aio app run --local --verbose` (openwhisk)

shazron:
Two changes:
1. The OpenWhisk jar URL has been updated to include the [Windows Docker fix](https://github.com/apache/openwhisk/pull/5021): https://github.com/adobe/openwhisk/releases/tag/standalone-v1
2. The OpenWhisk jar is not downloaded into the aio-cli-plugin-app's `bin` folder anymore, but in the cli's `dataDir` folder as specified in the [oclif config](https://oclif.io/docs/config). On macOS, this location for `@adobe/aio-cli` is ` ~/.local/share/@adobe/aio-cli/openwhisk/openwhisk-standalone.jar`. This is so the .jar is not deleted on every app plugin update.

## How Has This Been Tested?

npm test
aio app run --local --verbose

## Types of changes

## Checklist:

### DIFF
--- a/src/BaseCommand.js
+++ b/src/BaseCommand.js
@@ -20,8 +20,8 @@ class BaseCommand extends Command {
   getAppConfig () {
     if (!this.appConfig) {
       this.appConfig = loadConfig()
-      // add on app.cliConfig
-      this.appConfig.app.cliConfig = this.config
+      // add on appConfig
+      this.appConfig.cli = this.config
     }
     return this.appConfig
   }
--- a/src/BaseCommand.js
+++ b/src/BaseCommand.js
@@ -11,7 +11,6 @@ governing permissions and limitations under the License.
 */
 
 const { Command, flags } = require('@oclif/command')
-const fs = require('fs-extra')
 const chalk = require('chalk')
 const coreConfig = require('@adobe/aio-lib-core-config')
 const DEFAULT_LAUNCH_PREFIX = 'https://experience.adobe.com/?devMode=true#/custom-apps/?localDevUrl='
@@ -21,6 +20,8 @@ class BaseCommand extends Command {
   getAppConfig () {
     if (!this.appConfig) {
       this.appConfig = loadConfig()
+      // add on app.cliConfig
+      this.appConfig.app.cliConfig = this.config
     }
     return this.appConfig
   }
@@ -44,10 +45,7 @@ class BaseCommand extends Command {
   }
 
   get pjson () {
-    if (!this._pjson) {
-      this._pjson = fs.readJSONSync('package.json')
-    }
-    return this._pjson
+    return this.config.pjson
   }
 
   get appName () {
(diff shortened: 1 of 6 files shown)


src/BaseCommand.js
```diff
@@ -21,6 +20,8 @@ class BaseCommand extends Command {
   getAppConfig () {
     if (!this.appConfig) {
       this.appConfig = loadConfig()
+      // add on app.cliConfig
+      this.appConfig.app.cliConfig = this.config
```
moritzraho, line 24: I feel it's odd having the `cliConfig` inside the app config, should we use`this.appConfig.cli` instead ?

src/lib/runDev.js
```diff
@@ -44,6 +44,7 @@ const eventPoller = new EventPoller(fetchLogInterval)
 async function runDevLocal (config, cleanup, log) {
   const devConfig = cloneDeep(config)
   devConfig.envFile = path.join(config.app.dist, '.env.local')
+  const owJarFile = path.join(config.app.cliConfig.dataDir, 'openwhisk', 'openwhisk-standalone.jar')
```
moritzraho, line 47: big +1 to reusing the same DIR across CLI updates
However, if I update a new version of the app plugin that relies on a new `openwhisk-standalone.jar` will it download the new jar, or still use the old one because it's already there?
If it still uses the old one I guess we need to version the file names or use a md5 to force the jar update

moritzraho (commented):
2 inline comments

src/BaseCommand.js
shazron: Good point - I'll rename

src/lib/runDev.js
shazron: This is true, the download location on disk is not determinant on the URL that might have 
changed. 

Two options:
1. We could take the last part of the URL path? `openwhisk/standalone-v1/openwhisk-standalone.jar`. If we stick to this format, deriving this will be easy.
2. We could hash the URL, and use it for the local path. e.g. https://bintray.com/api/ui/download/adobe/generic/openwhisk/standalone-v1/openwhisk-standalone.jar could hash to `acde0a`, thus the path can be `openwhisk/acde0a/openwhisk-standalone.jar`

moritzraho: I like 1 ! it's readable in case the user wants to look at this.
Should we also clean up past versions (or open an issue and address this later if we get many more versions) ?

shazron: I think we should revisit if it gets to be a problem. Perhaps a new command to list and clean (not sure).

moritzraho: ok makes sense 🙂

shazron: Latest commit should resolve the two comments above.

moritzraho (approved):
thanks for the changes, lgtm

Issue #269, opened 2020-07-15
Enhancement: Add a tool to verify .env file


tmathern:
**Is your feature request related to a problem? Please describe.**
It is easily to oversee that en environment variables in the `.env` file was defined twice when debugging. 
The last definition wins, but sometimes it is not obvious because some variables are not case sensitive.
That is a burden for people discovering the tooling, since it is easily misconfigured and there is no warning or error.

**Describe the solution you'd like**
Either a tool to run which spits outs warning for variables defined twice, or integrate it in app run or app deploy (still displaying a warning for double definition).

**Describe alternatives you've considered**
Alternatives are to be careful while debugging. But seeing the duplicate value while debugging can be hard (especially if there is different casing).

**Additional context**
Found while supporting people during Nui Hackathon.

aiojbot:
JIRA issue created: https://jira.corp.adobe.com/browse/ACNA-854

meryllblanchet:
Assigning to @sarahxxu for evaluation.

PR #337, opened 2020-11-27
.env validation for required vars, duplicates and regex matching

Himavanth:
## Description
- Checking `.env` for duplicates https://github.com/adobe/aio-cli-plugin-app/issues/269
- Existence of a duplicate will result in a message in the debug logs that can be seen using `DEBUG=@adobe/aio-cli-plugin-app* aio app logs`
- Looking up in `.env`, cmd line args and process.env for required vars as defined in `.env.schema` if present
- Look up is case insensitive
- Validation of the variable values against regex if defined in `.env.schema`
- Schema validation failure results in an error
- Pending documentation

## Types of changes

## Checklist:

### DIFF
--- a/src/BaseCommand.js
+++ b/src/BaseCommand.js
@@ -17,9 +17,9 @@ const DEFAULT_LAUNCH_PREFIX = 'https://experience.adobe.com/?devMode=true#/custo
 const loadConfig = require('./lib/config-loader')
 
 class BaseCommand extends Command {
-  getAppConfig () {
+  getAppConfig (validateSchema) {
     if (!this.appConfig) {
-      this.appConfig = loadConfig()
+      this.appConfig = loadConfig(validateSchema)
       // add on appConfig
       this.appConfig.cli = this.config
     }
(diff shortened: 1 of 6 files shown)


Himavanth:
The technical debt argument on addition of a new file `.env.schema` by default through the default generator is a valid one.
While the tech debt cannot be nullified while adding a feature, I can think of two possible solutions here to minimize it.
1. Rename it to `.app.schema` since we are doing this only for the app plugin. It would still be a debt but only for the app plugin.
2. Do not put the file during app creation and make it's name configurable in `.env` via a variable say `AIO_app_schemafile`.
WDYT @purplecabbage @meryllblanchet @sarahxxu

meryllblanchet:
Hi @Himavanth ,

I'd be for keeping things simple for the time being, especially that the initial request was about detection of duplicates.

What about dropping of the regexp matching for now? @sarahxxu wdyt?

Himavanth:
We are now only doing the duplicate check. Since that is more common and we are only logging debug messages in case of duplicates, this PR will be closed in favour of https://github.com/adobe/aio-lib-core-config/pull/50.

PR #353, opened 2020-12-22
fix: write default app config on app init

### LINKED ISSUE #355: aio app init -y doesn't create the .aio file
Running `aio app init -y`  doesn't create the `.aio` file

sandeep-paliwal:
Always create .aio file with default app config for app init command.
Fixes #355 

## Description

When aio app is initialised with -y flag, .aio file was not created. 
Fixed this by writing .aio file with default app config at app init. When import or CLI login is used, .aio file will be merged with project config.
Default app config includes, name of actions, dist and web-src folder. This will help users to customise app if required.

## How Has This Been Tested?

unit tests, manual app init test

## Types of changes

## Checklist:

### DIFF
--- a/src/lib/import.js
+++ b/src/lib/import.js
@@ -447,7 +447,6 @@ async function writeConsoleConfig (json) {
  * @returns {Promise} promise from writeFile call
  */
 async function writeAio (json, parentFolder, flags) {
-  console.log('-----------writting aio-------------------------')
   aioLogger.debug(`writeAio - parentFolder:${parentFolder} flags:${flags}`)
   aioLogger.debug(`writeAio - json: ${prettyPrintJson(json)}`)
 
@@ -455,7 +454,6 @@ async function writeAio (json, parentFolder, flags) {
   aioLogger.debug(`writeAio - destination: ${destination}`)
 
   const data = prettyPrintJson(json)
-  console.log('-----------done-------------------------')
   return writeFile(destination, data, flags)
 }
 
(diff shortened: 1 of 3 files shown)


src/lib/import.js
```diff
@@ -95,6 +95,28 @@ function loadAndValidateConfigFile (fileOrBuffer) {
...
+ * @param {string} parentDir the parent folder to write the .aio file to
+ * @param {object} [flags] flags for file writing
+ * @param {boolean} [flags.overwrite=false] set to true to overwrite the existing .env file
+ * @param {boolean} [flags.merge=false] set to true to merge in the existing .env file (takes precedence over overwrite)
+ * @param {boolean} [flags.interactive=false] set to true to prompt the user for file overwrite
+ * @returns {Promise} promise from writeFile call
+ */
+function writeDefaultAppConfig (parentDir, flags) {
+  // write app config to .aio file
+  const appConfig = {
```
shazron, line 110: as an output, this has to be tested that it doesn't change unexpectedly

shazron (changes requested):

src/lib/import.js
sandeep-paliwal: @shazron there is test for `writeDefaultAppConfig ` in import.test.js to check output

shazron: Saw the test. I was thinking more like the contents here:
```
app: {
       actions: 'actions',
       dist: 'dist',
       web: 'web-src'
     }
```
A user might be relying on those values we output, so we shouldn't delete or change them (without a major version release) and a test would ensure this contract

sandeep-paliwal: Added the check to verify default aio content.

shazron (approved):

PR #419, opened 2021-06-29
fix: redefinition of command issue, pkg undefined

shazron:


shazron:
fixed in the target branch already

PR #421, opened 2021-06-29
Additional ext reg changes - ext and ext point listing

sandeep-paliwal:
Change impl prompt listing to dynamic, fetching extension point list from console
Add generic list command
Add list extension command to display implemented extensions and related details
and extension-points command to list xp available for the org
Added unit tests

## How Has This Been Tested?

manual tests

## Types of changes

## Checklist:

### DIFF
--- a/src/commands/app/add/extension.js
+++ b/src/commands/app/add/extension.js
@@ -15,9 +15,8 @@ const aioLogger = require('@adobe/aio-lib-core-logging')('@adobe/aio-cli-plugin-
 const { flags } = require('@oclif/command')
 const ora = require('ora')
 
-const { installPackages, atLeastOne } = require('../../../lib/app-helper')
+const { installPackages, atLeastOne, getImplPromptChoices } = require('../../../lib/app-helper')
 const aioConfigLoader = require('@adobe/aio-lib-core-config')
-const { implPromptChoices } = require('../../../lib/defaults')
 const chalk = require('chalk')
 
 class AddExtensionCommand extends BaseCommand {
@@ -59,7 +58,8 @@ class AddExtensionCommand extends BaseCommand {
 
   async selectImplementations (flags, config) {
     const alreadyImplemented = config.implements
-    const availableChoices = implPromptChoices
+    const consoleCLI = await this.getLibConsoleCLI()
+    const availableChoices = await getImplPromptChoices(consoleCLI)
     const availableImplementations = availableChoices.map(i => i.value.name)
 
     const possibleChoices = availableChoices.filter(i => !alreadyImplemented.includes(i.value.name))
--- a/src/commands/app/add/extension.js
+++ b/src/commands/app/add/extension.js
@@ -59,7 +59,7 @@ class AddExtensionCommand extends BaseCommand {
   async selectImplementations (flags, config) {
     const alreadyImplemented = config.implements
     const consoleCLI = await this.getLibConsoleCLI()
-    const availableChoices = await getImplPromptChoices(consoleCLI)
+    const availableChoices = await getImplPromptChoices(consoleCLI, config.aio)
     const availableImplementations = availableChoices.map(i => i.value.name)
 
     const possibleChoices = availableChoices.filter(i => !alreadyImplemented.includes(i.value.name))
--- a/src/commands/app/add/extension.js
+++ b/src/commands/app/add/extension.js
@@ -15,8 +15,9 @@ const aioLogger = require('@adobe/aio-lib-core-logging')('@adobe/aio-cli-plugin-
 const { flags } = require('@oclif/command')
 const ora = require('ora')
 
-const { installPackages, atLeastOne, getImplPromptChoices } = require('../../../lib/app-helper')
+const { installPackages, atLeastOne } = require('../../../lib/app-helper')
 const aioConfigLoader = require('@adobe/aio-lib-core-config')
+const { implPromptChoices } = require('../../../lib/defaults')
 const chalk = require('chalk')
 
 class AddExtensionCommand extends BaseCommand {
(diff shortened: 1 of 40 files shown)


src/lib/app-helper.js
```diff
@@ -516,6 +517,32 @@ function deleteUserConfig (configData) {
   fs.writeFileSync(configData.file, yaml.safeDump(phyConfig))
 }
 
+/** Get ALL Extension Point List for the selected org
+ *
+ * @param consoleCLI
+ */
+async function getAllExtensionPoints (consoleCLI) {
+  const projectConfig = aioConfig.get('project')
```
moritzraho, line 525: can you pass the aioConfig as param, it's available under the fullConfig.aio, that way we keep the loading of aioConfig inside the config-loader, that makes it easier for testing/mocking too

src/lib/app-helper.js
```diff
@@ -516,6 +517,32 @@ function deleteUserConfig (configData) {
...
+  return extensionPoints
+}
+
+/** Get promt choices for app
+ *
+ * @param consoleCLI
+ */
+async function getImplPromptChoices (consoleCLI) {
+  const defaultList = implPromptChoices
+  const projectConfig = aioConfig.get('project')
```
moritzraho, line 536: same comment here

moritzraho (commented):
Nice, good job integrating with new changes and generator-aio-console, especially because it's all ongoing work.

Just one double minor comment, otherwise lgtm !

src/lib/app-helper.js
sandeep-paliwal: I have changed the function interface to accept orgId( don't need entire config anyway)

src/lib/defaults.js
```diff
@@ -36,33 +36,39 @@ module.exports = {
...
+    'dx/asset-compute/worker/1': {
       name: 'DX Asset Compute Worker v1',
       value: {
         name: 'dx/asset-compute/worker/1',
         generator: generators.extensions['dx/asset-compute/worker/1'],
         requiredServices: ['AssetComputeSDK']
       }
     }
+  },
+  implPromptChoices: [
```
moritzraho, line 63: This is not very clean, implPromptChoices should include both extensions and standalone app, otherwise please rename that default

src/commands/app/add/extension.js
```diff
@@ -59,7 +58,8 @@ class AddExtensionCommand extends BaseCommand {
 
   async selectImplementations (flags, config) {
     const alreadyImplemented = config.implements
-    const availableChoices = implPromptChoices
+    const consoleCLI = await this.getLibConsoleCLI()
+    const availableChoices = await getImplPromptChoices(consoleCLI, config.aio)
```
moritzraho, line 62: I thought we decided that add extension and init should let the user choose extension points to implement from an hardcoded list, for two reasons:

- there is no added benefit of adding an extension from a dynamic list for now because we anyway maintain the list in the CLI
- dynamic list requires an extra request, slows down + more chance for bugs

moritzraho: cc @meryllblanchet @purplecabbage @sarahxxu

src/commands/app/init.js
```diff
@@ -108,7 +106,7 @@ class InitCommand extends BaseCommand {
     // 4. retrieve workspace details, defaults to Stage
     const workspace = await this.retrieveWorkspaceFromName(consoleCLI, org, project, flags.workspace)
     // 5. ask for exensionPoints, only allow selection for extensions that have services enabled in Org
-    const extensionPoints = await this.selectExtensionPoints(flags, orgSupportedServices)
+    const extensionPoints = await this.selectExtensionPoints(flags, orgSupportedServices, { login: true, orgId: org.id })
```
moritzraho, line 109: In case we keep dynamic extension list, please split this into two methods:
`this.selectExtensionPointsStatic(flags, orgSupportedServices)`
`this.selectExtensionPointsDynamic(consoleCLI, flags, orgSupportedServices, orgId)`

src/commands/app/init.js
```diff
@@ -132,12 +130,17 @@ class InitCommand extends BaseCommand {
     this.log(chalk.blue(chalk.bold(`Project initialized for Workspace ${workspace.name}, you can run 'aio app use -w <workspace>' to switch workspace.`)))
   }
 
-  async selectExtensionPoints (flags, orgSupportedServices = null) {
+  async selectExtensionPoints (flags, orgSupportedServices = null, options = {}) {
+    let consoleCLI
+    if (options.login) {
```
moritzraho, line 135: let's pass consoleCLI as input see comment above

src/commands/app/list/extension.js
```diff
@@ -0,0 +1,108 @@
...
+
+class ListExtensionCommand extends BaseCommand {
+  async run () {
+    const { flags } = this.parse(ListExtensionCommand)
+    aioLogger.debug(`list extensions with flags: ${JSON.stringify(flags)}`)
+
+    const extConfig = this.getAppExtConfigs(flags)
+    const extSummary = {}
+
+    EXTENSION_POINT_LIST.forEach(extPoint => {
```
moritzraho, line 29: This is half dynamic half static.
I think in this case we can go full dynamic and get extension points too ?

moritzraho (changes requested):
Comments below + **missing tests** for init, app-helper and add extension changes

src/commands/app/list/extension.js
```diff
@@ -0,0 +1,107 @@
...
+            {
+              impl: extension.operations.view[0].impl,
+              src: extension.web.src
+            }
+          ]
+        }
+        // get worker impl details
+        if (extension.operations.apply) {
+          // TODO extension.manifest.full.packages[extPoint.toString()] doesnt fetch package details
+          const pkgDetails = extension.manifest.full.packages['dx-asset-compute-worker-1']
```
moritzraho, line 47: that is not ok, the package name can be another.. why don't you simply print all operations, no need to filter on operation names ?

moritzraho (approved):
lgtm

Issue #468, opened 2021-08-25
App run auto watch/re-deployment of actions redeploys all


moritzraho:
## Current behavior

1. ```
> aio app run
```

2. modify an action file
3. all actions get redeployed

## Expected behavior

Should redeploy only actions for modified files


Relates to #409 
cc @mhaack @duynguyen @Himavanth

PR #478, opened 2021-09-16
[ACNA-1304] Filter actions deployments. 

florind12:
Changing the way deployments go.
On file change, it should deploy only its respective action.
It should not redeploy twice the same action packages if redeployment is called unless files were changed.

**Clean up after the e2e/unit test gets implemented.** 

## Description

Create a filter array based on the current file changed. This will help the deploy function to decide which actions to deploy. 

## Related Issue
https://github.com/adobe/aio-cli-plugin-app/issues/468

## Types of changes

## Checklist:

### DIFF
--- a/src/lib/actions-watcher.js
+++ b/src/lib/actions-watcher.js
@@ -75,30 +75,67 @@ async function buildAndDeploy (watcherOptions) {
 function createChangeHandler (watcherOptions) {
   const { watcher, log } = watcherOptions
 
-  let running = false
-  let changed = false
+  let deploymentInProgress = false
+  let fileChanged = false
+  let undeployedFile = ''
 
   return async (filePath) => {
-    if (running) {
+    const actionName = filePath.split('/')[filePath.split('/').findIndex(el => el === 'actions') + 1]
+    console.log(
+      '\n \n \n',
+      '\n Action file changed, ' + actionName + '\n',
+      '\n Path: ' + filePath + '\n',
+      'States::::::::::::::::::::::::::::::::::::::::::::\n',
+      '-> deploymentInProgress state::', deploymentInProgress, '\n',
+      '-> File changed during deploy state::', fileChanged, '\n',
+      '-> Last undeployedFile path was', undeployedFile
+    )
+    console.log('----> Starting the deployment for ', actionName)
+    if (deploymentInProgress) {
+      console.log(`----> another deploy is in progress for ${undeployedFile}, skipping for now`)
       aioLogger.debug(`${filePath} has changed. Deploy in progress. This change will be deployed after completion of current deployment.`)
-      changed = true
+      undeployedFile = filePath
+      fileChanged = true
       return
     }
-    running = true
+    deploymentInProgress = true
     try {
       aioLogger.debug(`${filePath} has changed. Redeploying actions.`)
+      const actionName = getActionNameFromPath(filePath)
+      watcherOptions.config.filterActions = [actionName]
       await buildAndDeploy(watcherOptions)
-      aioLogger.debug('Deployment successful.')
+      aioLogger.debug(`Deployment successful for', ${actionName}.`)
+      console.log('----> Deployment successful for ', actionName)
     } catch (err) {
+      console.log('----> something went wrong')
       log('  -> Error encountered while deploying actions. Stopping auto refresh.')
       aioLogger.debug(err)
       await watcher.close()
     }
-    if (changed) {
+    if (fileChanged) {
+      console.log('----> Code changed during deployment, Triggering deploy again for', actionName)
       aioLogger.debug('Code changed during deployment. Triggering deploy again.')
-      changed = running = false
-      await createChangeHandler(watcherOptions)(filePath)
+      fileChanged = deploymentInProgress = false
+      await createChangeHandler(watcherOptions)(undeployedFile)
     }
-    running = false
+    deploymentInProgress = false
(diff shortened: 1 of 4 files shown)


src/lib/actions-watcher.js
```diff
@@ -75,30 +75,67 @@ async function buildAndDeploy (watcherOptions) {
...
+ * @param {string} filePath  path of the file
+ * @returns {string}  name of the action
+ */
+function getActionNameFromPath (filePath) {
+  const defaultReturn = ''
+  if (!filePath) return defaultReturn
+  const pathArray = filePath.split('/')
+  /* validate the path */
+  if (pathArray.includes('actions')) {
+    const actionParentIndex = pathArray.findIndex(el => el === 'actions')
```
purplecabbage, line 138: pathArray.indexOf('actions') //??

florind12: Oh, right, missed that. 
Actually,  think this check is not necessary at all since the watcher is over the root of the actions files.

src/lib/actions-watcher.js
```diff
@@ -75,30 +75,46 @@ async function buildAndDeploy (watcherOptions) {
...
   }
 }
+
+/**
+ * Util function which extract the actionName from the filePath.
+ *
+ * @param {string} filePath  path of the file
+ * @returns {string}  name of the action
+ */
+function getActionNameFromPath (filePath) {
```
moritzraho, line 116: that won't always work unfortunately, the action name does not have to be related to the path, you will have to match the path to the action name from the info in the manifest, and even then beware, the manifest can define multiple actions for the same action path (with different inputs for example)

florind12: Hmm, having multiple actions with the same action path would be a hassle.
I suppose this approach won't work at all.

florind12: "the manifest can define multiple actions for the same action path (with different inputs for example)"

@moritzraho Any idea where could I find such an example?

moritzraho: Let's assume the following manifest actions
```
actions:
  hello:
     function: 'my/path/to/single-action.js'
     input:
       default: 'param'
  hello-2:
     function: 'my/path/to/single-action.js'
     input:
       default: 'paramvalue2'
```


If `'my/path/to/single-action.js'` changes then you will have to redeploy the two actions based on the manifest information

florind12: In this case, I would return an array with all of the matches.

src/lib/actions-watcher.js
```diff
@@ -75,30 +75,58 @@ async function buildAndDeploy (watcherOptions) {
...
+      undeployedFile = filePath
+      fileChanged = true
       return
     }
-    running = true
+    deploymentInProgress = true
     try {
       aioLogger.debug(`${filePath} has changed. Redeploying actions.`)
+      const actionNames = getActionNameFromPath(filePath, watcherOptions)
+      if (actionNames.length) {
```
moritzraho, line 93: let's do `actionNames.length > 0`

src/lib/deploy-actions.js
```diff
@@ -25,7 +25,15 @@ module.exports = async (config, isLocal = false, log = () => {}) => {
   utils.runScript(config.hooks['pre-app-deploy'])
   const script = await utils.runScript(config.hooks['deploy-actions'])
   if (!script) {
-    const entities = await deployActions(config, { isLocalDev: isLocal }, log)
+    const deployConfig = {
+      isLocalDev: isLocal
+    }
+    if (config.filterByBuiltActions) {
+      deployConfig.filterEntities = {}
```
moritzraho, line 32: please don't add fields to the config object, it has a specific format as defined by the config-loader, I don't think we should mess up with this spec. Any way to pass this as an additional argument ?

src/lib/run-dev.js
```diff
@@ -122,6 +122,7 @@ async function runDev (config, dataDir, options = {}, log = () => {}) {
     // Deploy Phase - deploy actions
     if (withBackend) {
       log('redeploying actions..')
+      devConfig.filterByBuiltActions = true
```
moritzraho, line 125: same comment here, please don't overload the config object

.gitignore
```diff
@@ -14,3 +14,11 @@ oclif.manifest.json
 package-lock.json
 
 bin/openwhisk-standalone*.jar
+/.idea/.gitignore
```
moritzraho, line 17: what about just /.idea/ ?

moritzraho (approved):
few nitpicks comment otherwise lgtm, feel free to merge after resolving those

src/lib/actions-watcher.js
```diff
@@ -75,30 +76,55 @@ async function buildAndDeploy (watcherOptions) {
...
+ * @param {WatcherOptions} watcherOptions the options for the watcher
+ * @returns {Array<string>}  All of the actions which match the modified path
+ */
+function getActionNameFromPath (filePath, watcherOptions) {
+  const actionNames = []
+  const { config } = watcherOptions
+  Object.entries(config.manifest.full.packages).forEach(([, pkg]) => {
+    if (pkg.actions) {
+      Object.entries(pkg.actions).forEach(([actionName, action]) => {
+        if (action.function.includes(filePath)) {
```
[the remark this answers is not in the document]
florind12: Yep, you are right, `action.function` will have a Unix path even if it's generated on windows.

Issue #377, opened 2021-02-10
Support workspace creation in aio app use ?


moritzraho:
## Current Behavior
currently we are allowed to switch to a different workspace in the project, via a prompt: 
```
You are currently in:
1. Org: <>
2. Project: <>
3. Workspace: <>

? Switch to a new Adobe Developer Console configuration: B. Switch to another Workspace in the current Project
? Select Workspace: (Use arrow keys or type to search)
❯ Production
  Stage
```

Or via a flag `aio app use -w workspaceName`, in this case we check that the workspace exists

## Enhancement
In addition we could allow the creation of a new workspace to switch too, we would also benefit from the existing sync services logic.

Like in `aio app init` the prompt could look like:
```
? Select a Workspace, or press + to create new: (Use arrow keys or type to search)
❯ Production
  Stage
```
pressing `+` allows to create a new workspace

The flag version of it could be `aio app use -w workspaceName --new-workspace`

aiojbot:
JIRA issue created: https://jira.corp.adobe.com/browse/ACNA-1056

PR #498, opened 2021-10-20
[Feat.][ACNA-1056] Support workspace creation in aio app use -w and aio app init -w

florind12:
## Description
Switching to a different workspace currently is done via the prompt or via flag. 
In addition to this, we want to be able to create and switch to a new workspace as well.
`aio app use` should support workspace creation via prompt.
`aio app use --workspace` should support workspace creation in case the requested workspace doesn't exist.
`aio app use --workspace --confirm-new-workspace` would skip prompt the confirmation on the workspace creation process.
`aio app use --workspaceId` should switch to the workspace by the given id.

## Related Issue
https://github.com/adobe/aio-cli-plugin-app/issues/377

## Types of changes

## Checklist:


src/commands/app/use.js
```diff
@@ -190,37 +190,84 @@ class Use extends BaseCommand {
...
+      workspace = workspacesButCurrent.find(currentWorkspace => this.workspaceExist(currentWorkspace, workspaceDetails))
+      if (!workspace && workspaceNameFlag) {
+        aioLogger.debug(`--workspace=${workspaceNameOrId} was not found in the current Project ${project.name}.
+Creating workspace ${workspaceNameOrId}...`)
+        const skipPrompt = !!flags['confirm-new-workspace']
+        // 2.1. Create the workspace with the given name.
+        if (!skipPrompt) {
+          const shouldNewWorkspace = await consoleCLI.prompt.promptConfirm(`Workspace ${workspaceNameFlag} was not found \n > Do you wish to create a new workspace named: '${workspaceNameFlag}'`)
+          if (!shouldNewWorkspace) {
+            this.error(`Workspace creation for ${workspaceNameFlag} was denied by user`)
```
moritzraho, line 230: nitpick: `Workspace creation aborted`

src/commands/app/use.js
```diff
@@ -190,37 +190,84 @@ class Use extends BaseCommand {
...
+    if (workspaceNameOrId) {
+      // 2. Find a workspace with the given name or id
+      workspace = workspacesButCurrent.find(currentWorkspace => this.workspaceExist(currentWorkspace, workspaceDetails))
+      if (!workspace && workspaceNameFlag) {
+        aioLogger.debug(`--workspace=${workspaceNameOrId} was not found in the current Project ${project.name}.
+Creating workspace ${workspaceNameOrId}...`)
+        const skipPrompt = !!flags['confirm-new-workspace']
+        // 2.1. Create the workspace with the given name.
+        if (!skipPrompt) {
+          const shouldNewWorkspace = await consoleCLI.prompt.promptConfirm(`Workspace ${workspaceNameFlag} was not found \n > Do you wish to create a new workspace named: '${workspaceNameFlag}'`)
```
moritzraho, line 228: nitpick: `Workspace '${workspaceNameOrIdFlag}' does not exist \n > Do you wish to create a new workspace ?`

moritzraho (changes requested):
Some comments +

I think we can simplify all the creation logic by supporting one change in aio-cli-lib-console, if in https://github.com/adobe/aio-cli-lib-console/blob/master/lib/index.js#L125 you return `undefined` instead of throwing on non existing when allow-create is true then you can combine the cases where the workspace flag is set or when the user is prompted

Then the code could be simplified to something like:
```javascript
const workspaceNameOrIdFlag = flags.workspace
const workspaces = await consoleCLI.getWorkspaces(org.id, project.id)
let workspace = await consoleCLI.promptForSelectWorkspace(workspaces, {workspaceId: workspaceNameOrIdFlag, workspaceName: workspaceNameOrIdFlag },{ 'allow-create': true })
if (!workspace) {
// we need to create
   if (workspaceNameOrIdFlag && !flags['confirm-new-workspace']) {
          // confirm creation in for workspace flag and not confirm flag (this could even be moved to aio-cli-lilb-consolle
          const confirmNewWorkspace = await consoleCLI.prompt.promptConfirm(`Workspace '${workspaceNameOrIdFlag}' does not exist \n > Do you wish to create a new workspace ?`)
          if (confirmNewWorkspace) {
            this.error(`Workspace creation for ${workspaceNameFlag} was denied by user`)
          }
    workspaceDetails = { name: workspaceNameOrIdFlag }
  }
  else {
      const workspaceDetails = await consoleCLI.promptForCreateWorkspaceDetails()
  }
    await consoleCLI.createWorkspace(orgId, projectId, workspaceDetails)
}
```

src/commands/app/init.js
```diff
@@ -189,9 +189,10 @@ class InitCommand extends AddCommand {
   async retrieveWorkspaceFromName (consoleCLI, org, project, workspaceName) {
     // get workspace details
     const workspaces = await consoleCLI.getWorkspaces(org.id, project.id)
-    const workspace = workspaces.find(w => w.name.toLowerCase() === workspaceName.toLowerCase())
+    let workspace = workspaces.find(w => w.name.toLowerCase() === workspaceName.toLowerCase())
     if (!workspace) {
-      throw new Error(`'--workspace=${workspaceName}' in Project '${project.name}' not found.`)
+      this.log(`'--workspace=${workspaceName}' in Project '${project.name}' not found. \n Creating one...`)
```
purplecabbage, line 194: We should ask the user if they want to create a new workspace titled `workspaceName`, blindly creating it when a user might have simply mistyped seems like overkill.

README.md
```diff
@@ -780,11 +780,13 @@ OPTIONS
 
   -v, --verbose                        Verbose output
 
-  -w, --workspace=workspace            Specify the Adobe Developer Console Workspace name to import the configuration
-                                       from
+  -w, --workspace=workspace            Specify the Adobe Developer Console Workspace name or Workspace id to import the
```
purplecabbage, line 783: We should not be manually updating the readme, this is a generated file at the time of an npm publish ... otherwise we can expect merge issues to happen.

Issue #501, opened 2021-11-18
Add commands for managing log forwarding


buskamuza:
**Is your feature request related to a problem? Please describe.**

I/O Runtime added an ability to configure log forwarding, so that activation logs are forwarded to a custom destination. Supported platforms: Splunk and Azure Log Analytics.
This is a request to add `aio app` commands to manage log forwarding for the namespace where the app is being deployed.
See https://github.com/adobe/aio-cli-plugin-runtime/issues/247 , which introduces same functionality for `aio rt` commands.

**Describe the solution you'd like**

Expected commands:

```
aio app:add:config:get:log-forwarding - get current log forwarding settings
aio app:add:config:set:log-forwarding - interactively set log forwarding configuration
aio app:add:config:get:log-forwarding - fetch log forwarding errors, if any
```

Supported log forwarding destinations are based on what is provided by Runtime API (https://github.com/adobe/aio-lib-runtime). So, as new destinations are added the command should automatically support those.

If the log forwarding configuration is set to `adobe_io_runtime` (default), `aio app:logs` continue to work as they do today.
If the log forwarding configuration is set to a custom destination, then `aio app:logs` should display a helpful message to the developer.

**Describe alternatives you've considered**
N/A

PR #502, opened 2021-11-18
App Builder experience for Configurable Logging

buskamuza:
Added commands for managing log forwarding configuration.

- added `app:config:get:log-forwarding` command
- added `app:config:set:log-forwarding` command
- added `app:config:get:log-forwarding:errors` command
- updated log forwarding config on server during `app:deploy` command
- updated `app:logs` command to display info message in case custom log forwarding is configured
- save log forwarding settings to project config

## Description

See https://github.com/adobe/aio-cli-plugin-app/issues/501 and https://jira.corp.adobe.com/browse/DEVX-1244 (internal)

## Related Issue

- https://github.com/adobe/aio-cli-plugin-app/issues/501
- https://github.com/adobe/aio-lib-runtime/pull/83
- https://github.com/adobe/aio-cli-plugin-runtime/pull/254
- https://github.com/adobe/aio-lib-runtime/pull/93

## Motivation and Context

See https://github.com/adobe/aio-cli-plugin-app/issues/501

## How Has This Been Tested?

Automated and manual testing.

## Types of changes

## Checklist:


src/lib/log-forwarding.js
```diff
@@ -0,0 +1,253 @@
...
+  }
+
+  async init () {
+    rtLib.utils.checkOpenWhiskCredentials({ ow: this.aioConfig.runtime })
+    this.logForwarding = await getRTLogForwarding(this.aioConfig.runtime)
+    return this
+  }
+
+  getLocalConfig () {
+    const config = this.aioConfig.project.workspace.log_forwarding
```
purplecabbage, line 33: can this be null?

src/lib/log-forwarding.js
```diff
@@ -0,0 +1,253 @@
...
+  let destination
+  let settings
+
+  if (configJson !== undefined) {
+    const destinations = Object.keys(configJson)
+    if (destinations.length === 1) {
+      destination = destinations[0]
+      settings = configJson[destination]
+    } else {
+      throw new Error(`Configuration has ${destinations.length} destinations. Exactly one must be defined.`)
```
purplecabbage, line 221: I am not sure why this check is required, how would we get a config object with multiple destinations?  Seems like that would be an error in code.

src/lib/log-forwarding.js
```diff
@@ -0,0 +1,253 @@
...
+    const config = this.getLocalConfig()
+    const destination = config.getDestination()
+    const settings = config.getSettings()
+    if (config.isDefined()) {
+      const destinationSettings = this.logForwarding.getDestinationSettings(destination)
+      const missingSecrets = []
+      destinationSettings.forEach(e => {
+        if (e.type === SECRET_FIELD_TYPE) {
+          const secretVarName = getSecretVarName(destination, e.name)
+          if (process.env[secretVarName] !== undefined) {
```
purplecabbage, line 51: this is very mysterious ... who sets this secret? how? why process.env?

src/lib/log-forwarding.js
```diff
@@ -0,0 +1,253 @@
...
+
+  async getServerConfig () {
+    try {
+      return convertToConfigObject(await this.logForwarding.get())
+    } catch (e) {
+      throw new Error('Incorrect log forwarding configuration on server. ' + e.message)
+    }
+  }
+
+  getSupportedDestinations () {
```
purplecabbage, line 74: nit: 1 line becomes 3 lines, just call this.logForwarding.getSupportedDestinations() directly

src/lib/log-forwarding.js
```diff
@@ -0,0 +1,253 @@
...
+    return this.logForwarding.getSupportedDestinations()
+  }
+
+  getSettingsConfig (destination) {
+    return this.logForwarding.getDestinationSettings(destination)
+  }
+
+  async updateLocalConfig (lfConfig) {
+    const destination = lfConfig.getDestination()
+    const destinationSettings = this.logForwarding.getDestinationSettings(destination)
```
purplecabbage, line 84: nit: we defined `getSettingsConfig` 5 lines ago ... why aren't we using it?
just kidding: this line is correct, I would remove the `getSettingsConfig` function all together, just like `getSupportedDestinations` above.

src/lib/log-forwarding.js
```diff
@@ -0,0 +1,253 @@
...
+
+  async updateServerConfig (lfConfig) {
+    await this.logForwarding.setDestination(lfConfig.getDestination(), lfConfig.getSettings())
+    const checksum = getChecksum(lfConfig)
+    fs.ensureDirSync(process.cwd() + '/tmp')
+    fs.writeFile(process.cwd() + '/tmp/' + CHECKSUM_FILE, checksum, { flags: 'w' })
+  }
+}
+
+class LogForwardingConfig {
```
purplecabbage, line 130: is this worth an object/class?

src/lib/log-forwarding.js
```diff
@@ -0,0 +1,253 @@
...
+      [destination]: nonSecretSettings
+    }
+    const interactive = false
+    const merge = true
+    await writeAio(projectConfig, process.cwd(), { interactive, merge })
+    await writeEnv({}, process.cwd(), { interactive, merge }, secretSettings)
+  }
+
+  isLocalConfigChanged () {
+    if (fs.pathExistsSync(process.cwd() + '/tmp/' + CHECKSUM_FILE)) {
```
purplecabbage, line 112: - Use the dist/ folder for this, which is part of aioConfig
- this code is not windows friendly either
- process.cwd() is not needed
- use path.join instead of adding path strings yourself

purplecabbage (changes requested):
mostly nitpicky stuff, but this seems to be a lot more code than required ... more code, more bugs

Deploy can be called with `--[no-]actions` in which case we should not be doing anything with log-forwarding.

src/lib/log-forwarding.js
buskamuza: If somebody writes it as `null` then yes. I'm checking for `undefined` now. But we can add more checks to take into account cases when people change the config manually and put something unexpected.

src/lib/log-forwarding.js
buskamuza: If somebody changes config manually. We can't guarantee this doesn't happen, so there should be intentional behavior in this case. But in normal scenario this shouldn't happen.

src/lib/log-forwarding.js
buskamuza: I looked at an examples like this one https://github.com/adobe/aio-lib-core-config/blob/master/src/Config.js#L51 
Is there a better way to read values from `.env` file?

buskamuza: Though I agree that this code depends on the `dotenv(true)` happening before it (which happens during config load). So if there is a better way (a clear API for reading secret data from `.env` or whatever place is expected for this data), let's use it.

src/lib/log-forwarding.js
buskamuza: Yes. I was thinking about those two ways. The only reason for the current implementation is fewer dependencies in the actual command. Right now, it depends only on this library. If we remove this method, it will have to depend also on the RT LogForwarding directly. I chose one of the options as it didn't look critical for me and this library looks just like a wrapper. But if it's a concern, I can remove these two methods.

src/lib/log-forwarding.js
buskamuza: Updated

src/lib/log-forwarding.js
buskamuza: IMO, client code (commands) look better now. But this is probably objective.

package.json
```diff
@@ -23,6 +23,7 @@
     "chalk": "^4.0.0",
     "chokidar": "^3.5.2",
     "cli-ux": "^5.4.5",
+    "crypto": "^1.0.1",
```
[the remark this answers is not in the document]
buskamuza: Removed

src/lib/log-forwarding.js
purplecabbage: you are checking for undefined, what I mean though is that this code will throw if workspace, or project, or aioConfig is null/undefined.
if we were node@14+ then you could use optional chaining 
`const config = this.aioConfig?.project?.workspace?.log_forwarding`

src/lib/log-forwarding.js
purplecabbage: my mistake, didn't see that this was called from outside

src/lib/log-forwarding.js
buskamuza: I haven't seen "parent" parts to be null and I checked with other code and it doesn't check these levels for null/undefined. But I can do it.

buskamuza:
> mostly nitpicky stuff, but this seems to be a lot more code than required ... more code, more bugs
> 
> Deploy can be called with `--[no-]actions` in which case we should not be doing anything with log-forwarding.

Updated

purplecabbage:
a couple issues I have found.
- when setting with `https://` the server does not return an `https://` url so `aio app config get log-forwarding` always shows a mismatch
- changing type via `aio app config set log-forwarding` overwrites values in .aio but leaves keys in .env

buskamuza:
> changing type via aio app config set log-forwarding overwrites values in .aio but leaves keys in .env

This was discussed with Manik. It leaves keys in `.env` if you change destination type. If you use the same destination type the key is overwritten. So far we agreed that this is expected behavior.

buskamuza:
> when setting with https:// the server does not return an https:// url so aio app config get log-forwarding always shows a mismatch

Fixed. Updated Runtime API to return sanitized settings and aio CLI then stores this instead of what the user provided. Secrets are not exchanged and stored as provided by the user. See also new PR https://github.com/adobe/aio-lib-runtime/pull/93 necessary to cover this.

PR #514, opened 2022-02-14
feat: add App Builder Template management commands

shazron:
## Description

Add app template commands. See `aio app template --help`
Support for non-npm registries are under the `--experimental-registry` flag under the `discover` command.

The experimental registry currently is at https://github.com/adobe/aio-app-builder-template-registry and you can use this [url](https://raw.githubusercontent.com/adobe/aio-app-builder-template-registry/main/data/metadata.json) for the `--experimental-registry` value.

## How Has This Been Tested?

Local link the plugin, and run the commands.

```bash
git clone https://github.com/adobe/aio-cli-plugin-app.git ACNA-1421
cd ACNA-1421
git checkout -b ACNA-1421 origin/ACNA-1421
npm i
aio plugins link .
# go to your App Builder app root, see available commands
aio app template --help
```

to unlink:
```bash
# go to your cloned repo root
cd ACNA-1421
aio plugins unlink .
```

## Types of changes

## Checklist:

- [X] I have updated the documentation accordingly.


package.json
```diff
@@ -59,6 +59,7 @@
     "eslint-plugin-promise": "^5.2.0",
     "eslint-plugin-standard": "^4.0.1",
     "jest": "^26.6.3",
+    "jest-fetch-mock": "^3.0.3",
```
shazron, line 62: note to self: can be removed

src/lib/npm-helper.js
```diff
@@ -0,0 +1,202 @@
...
+  const pjson = JSON.parse(fs.readFileSync(pjsonPath))
+
+  return pjson.version
+}
+
+/**
+ * Hide NPM Warnings by intercepting process.stderr.write stream
+ *
+ */
+function hideNPMWarnings () {
```
purplecabbage, line 173: This seems like a bad practice, and has side-effects.
Can't we just use --loglevel=error for the `npm install`

purplecabbage (commented):
This all looks good, however I think I would greatly prefer if we could split all `app template` commands into a separate `app-template` command, which still handled app:template commands identically, but this would allow us to pre-release / test this command independently.

src/lib/npm-helper.js
shazron: this was a copy paste from the discover plugin - to suppress warnings one level deeper that we have no control over (the plugin install command). Yes we should definitely change to what you recommended

shazron:
> This all looks good, however I think I would greatly prefer if we could split all `app template` commands into a separate `app-template` command, which still handled app:template commands identically, but this would allow us to pre-release / test this command independently.

yup this is the plan going forward now

shazron:
Closed with https://github.com/adobe/aio-cli-plugin-app-templates/pull/1

PR #515, opened 2022-02-21
fix: ACNA-1493 - pick up logs from __secure actions

### LINKED ISSUE #513: `aio app logs` does not pick up logs from __secure  actions
This is in opposition to `aio rt activation logs` which will get all the logs. Both commands call the same function in `@adobe/aio-lib-runtime`.

# Expected
`aio app logs` gets all logs for the App Builder app.

# Actual
`aio app logs` gets all logs, except from the __secure actions, for the App Builder app.

florind12:
## Description
`aio app logs`  should get all logs for the App Builder app.

 `rtLib.printActionLogs(..., filterActions, ...)` is called in order to print the logs.
filterActions param contains an array of actions to fetch logs from
__secured actions are left out since the filterActions array doesn't include them.

## Related Issue

Closes #513

## How Has This Been Tested?
unit tests.

## Screenshots (if appropriate):
image

## Types of changes

## Checklist:


Himavanth:
@florind12 We have a mechanism of fetching sub-activations from a sequence's activation at https://github.com/adobe/aio-lib-runtime/blob/master/src/utils.js#L394
I am guessing that is not working in this case because there is probably no sequence activation being returned from runtime. Could you check if this is normal or something changed recently ?
If this is the way forward for all cases, then we might have to remove the above logic to fetch action activations from a sequence activation.

shazron (changes requested):

shazron (approved):

PR #537, opened 2022-05-27
feat: ACNA-1602 - add new flag to prevent action code deployment

### LINKED ISSUE #538: Feature Request - Deploy changes to secrets from .env file without deploying the entire application
### Expected Behaviour
I update some values in my .env that are secrets needed by the application. When I deploy, I want an option to only deploy changes to the .env and not redeploy the entire application with code and actions and everything.

### Actual Behaviour
No way to only update .env parameters. It will redeploy the entire application along with code and actions and everything else.

### Reproduce Scenario (including but not limited to)
aio app deploy

#### Steps to Reproduce
1. Update secret parameter in .env
2. aio app deploy
3. it will redeploy the entire application, not just the .env updates which are the only things that have changed.

#### Platform and Version
@adobe/aio-cli/8.2.0 darwin-x64 node-v12.22.4

#### Sample Code that illustrates the problem
Any AppBuilder application

#### Logs taken while reproducing problem
aio app deploy logs

arjuncooliitr:
## Description

--[no-]action-code flag added to the aio-cli-plugin-app deploy command to prevent action code deployment

## Pre-requisite

- https://github.com/adobe/aio-lib-runtime/pull/98

## Related Issue

Closes #538

## Types of changes

## Checklist:


README.md
```diff
@@ -562,6 +562,8 @@ OPTIONS
 
   --web-optimize                [default: false] Enable optimization (minification) of web js/css/html
 
+  --[no-]code                   [default: true] --no-code prevents action code deployment
```
moritzraho, line 565: that's auto-generated no need to add it now ? right @shazron @purplecabbage ?

src/commands/app/deploy.js
```diff
@@ -337,6 +340,11 @@ Deploy.flags = {
     description: '[default: true] Update log forwarding configuration on server',
     default: true,
     allowNo: true
+  }),
+  code: flags.boolean({
+    description: '[default: true] --no-code prevents deploying action code',
```
moritzraho, line 345: - `--no-code` could be confusing for apps that have UI code too, maybe we should rename this to be more specific, what about `--no-action-code` or `--no-action-update`
- also should we add the flag to `app run` too ?
cc @purplecabbage @shazron

purplecabbage: I agree on the confusion factor.  I would much prefer if this was just another command.
Can we enumerate all the flag combos with all the expected operations?
```
aio app deploy --no-web-assets # actions and their params are deployed
aio app deploy --no-actions  # web-assets are deployed
aio app deploy --no-code     # web-assets and params are deployed
```
The pattern this is developing into is moving in the wrong direction imho.  I think we should be specifying what we _want_ to do, not what we don't want.

It might make more sense to leave this as just an `aio runtime` command ...

README.md
shazron: yes this is auto-generated right before publish (via the prepack npm script)

src/commands/app/deploy.js
shazron: I think specifying what we want to do instead of what we don't want is a re-factor decision. This would have been better for sure with (borrowing from Firebase) `--only functions` etc so we could have: `--only action-params` for example.

For now we have this pattern so this new flag has to fit in to that pattern. `--no-actions` means no action code and thus no params are deployed (thus for both). So to specify that action params are deployed, but *not* action code, I would say the flag is `--no-action-code`. Does deploying action code but not action params make sense? I don't think that the api allows for it anyway, or that we need to allow for it, so we won't need a flag for that scenario.

Expected to deploy:
| flag                             | actions | action params | web assets |
|----------------------------------|---------|---------------|------------|
| --no-actions                     | N       | N             | Y          |
| --no-action-code                 | N       | Y             | Y          |
| --no-web-assets                  | Y       | Y             | N          |
| --no-web-assets --no-action-code | N       | Y             | N          |

shazron (changes requested):
see threaded comments for changes

src/commands/app/deploy.js
shazron: This is the full truth table:
| flag                             | actions | action params | web assets |
|----------------------------------|---------|---------------|------------|
| --no-actions --no-web-assets     | N       | N             | N          |
| --no-actions                     | N       | N             | Y          |
| --no-action-code --no-web-assets | N       | Y             | N          |
| --no-action-code                 | N       | Y             | Y          |
| (invalid)                        | Y       | N             | N          |
| (invalid)                        | Y       | N             | Y          |
| (no flags)                       | Y       | Y             | Y          |

README.md
moritzraho: thanks for confirming

src/commands/app/deploy.js
arjuncooliitr: Following the discussion, I have changed the name of flag from `--no-code` to `--no-action-code` to avoid confusion with the UI code as stated by @moritzraho. The behavior of this flag is intended to be precisely as @shazron mentioned in the truth table.

shazron:
@arjuncooliitr 1 unit test failure - can you fix so we can add this for a potential Monday release?

shazron (approved):

purplecabbage:
It looks like it, but can you verify that these are supported?

Update the params for **action1** only ( assume there are many actions in the project )
`aio app deploy --no-web-assets -a action1 --no-action-code`

Update the params and deploy **action1** only ( many actions in project )
`aio app deploy --no-web-assets -a action1`

shazron:
@arjuncooliitr can you fix up the conflict? thanks!

arjuncooliitr:
@purplecabbage Sorry this has been long overdue 😅 but you mentioned long back that you had some thoughts regarding the semantics of this flag, so we didn't merge and kept it on hold. 
What are your thoughts now? If it looks okay to you then we can proceed and merge.

arjuncooliitr:
@shazron, resolved the conflict thanks!

shazron:
@arjuncooliitr can you resolve the conflict? 
@purplecabbage please see Arjun's comments above so this can be resolved.

PR #544, opened 2022-06-02
feat: ACNA-1678 - add `--force-build` flag for `aio app run`

### LINKED ISSUE #542: for `aio app run --local` (local deploy), if you ran `aio app run` (remote deploy) beforehand, it won't deploy the action code (and vice versa)
## Expected

Action code is deployed to the local Openwhisk server.

## Actual

Action code is not deployed to the local Openwhisk server, because the CLI thinks there are no changes.

## Investigation

When you ran `aio app run` previously (for remote deploy), it caches a hash of the built actions, and stores the values in `dist/last-built-actions.json`. This hash is checked, so if it is the same, it won't build and won't deploy the action.

However, for `aio app run --local` it is now a local deploy, and you haven't deployed yet (and it won't deploy because the file hashes are the same). So you will get an error during debugging that it can't find the action file.

This also happens vice versa.
## Fix

1. https://github.com/adobe/aio-cli-plugin-app/blob/4c4950f4d6cab65c62be56bd0ed7f3f307d05378/src/lib/build-actions.js#L22 add `skipCheck` as the third parameter, and pass it in as the third parameter here https://github.com/adobe/aio-cli-plugin-app/blob/4c4950f4d6cab65c62be56bd0ed7f3f307d05378/src/lib/build-actions.js#L26
2. https://github.com/adobe/aio-cli-plugin-app/blob/4c4950f4d6cab65c62be56bd0ed7f3f307d05378/src/lib/run-dev.js#L83 (pass in true as the third parameter. the second param set to an empty array)
3. Add a new flag to `aio app run`, like `aio app build`: `--force-build` which will always run the build (skipCheck=true above)

## Workaround

Delete the `dist/last-built-actions.json` file (or run `aio app build` which always forces a build), then run `aio app run --local`

shazron:
## Description

Fixes #542

## How Has This Been Tested?

npm test
manual test

## Types of changes

## Checklist:

- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly (will be auto-gen'ed)

### DIFF
--- a/src/commands/app/run.js
+++ b/src/commands/app/run.js
@@ -79,7 +79,8 @@ class Run extends BaseCommand {
       },
       fetchLogs: true,
       devRemote: !flags.local,
-      verbose: flags.verbose
+      verbose: flags.verbose,
+      forceBuild: flags['force-build']
     }
 
     try {
@@ -226,6 +227,11 @@ Run.flags = {
   open: flags.boolean({
     description: 'Open the default web browser after a successful run, only valid if your app has a front-end',
     default: false
+  }),
+  'force-build': flags.boolean({
+    description: '[default: false] Force a build even if one already exists',
+    default: false,
+    allowNo: true
   }),
   extension: flags.string({
     description: 'Run only a specific extension, this flag can only be specified once',
--- a/src/lib/build-actions.js
+++ b/src/lib/build-actions.js
@@ -18,12 +18,13 @@ const { buildActions } = require('@adobe/aio-lib-runtime')
  *
  * @param {object} config see src/lib/config-loader.js
  * @param {Array<string>} filterActions add filters to deploy only specified OpenWhisk actions
+ * @param {object} [forceBuild=false] set to true to always build (do not check if there were changes via hashing)
  */
-module.exports = async (config, filterActions) => {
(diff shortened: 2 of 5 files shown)


moritzraho (commented):
Is there a use case for force build ? can't we just force build transparently on local dev ?
To me force build sounds a bit like we might have some bugs "so just try force build" 😄

moritzraho:
> It's purely for the situation of switching between these two contexts.

Ok maybe we can avoid introducing the new flag by catching this case in code and "force build" transparently to the user instead ?

shazron:
> > It's purely for the situation of switching between these two contexts.
> 
> Ok maybe we can avoid introducing the new flag by catching this case in code and "force build" transparently to the user instead ?

Let me think about it some more. We would have to store the previous state somewhere (remote vs local).

shazron:
Now that I think about it some more, the workaround is just doing a `aio app build` (which always forces the build) before your aio app run if you switch contexts. `aio app build && aio app run --local`

shazron:
You've convinced me. We do an automatic force build on `aio app run` instead.

moritzraho:
hahaha ok ok

shazron:
Closing. Will do another PR for automatic force build on run.

PR #551, opened 2022-06-06
fix: force build on every aio app run (for switching from remote and local contexts)

### LINKED ISSUE #542: for `aio app run --local` (local deploy), if you ran `aio app run` (remote deploy) beforehand, it won't deploy the action code (and vice versa)

shazron:
## Description

Fixes #542 
(see issue for use case)

In a nutshell, we will always force deploy actions on every `aio app run` (doesn't matter if there are no code changes).

## How Has This Been Tested?

npm test
manual test

## Types of changes

## Checklist:

### DIFF
--- a/src/lib/build-actions.js
+++ b/src/lib/build-actions.js
@@ -18,12 +18,13 @@ const { buildActions } = require('@adobe/aio-lib-runtime')
  *
  * @param {object} config see src/lib/config-loader.js
  * @param {Array<string>} filterActions add filters to deploy only specified OpenWhisk actions
+ * @param {boolean} [forceBuild=false] force a build (skip file changed hash check)
  */
-module.exports = async (config, filterActions) => {
+module.exports = async (config, filterActions, forceBuild = false) => {
   utils.runScript(config.hooks['pre-app-build'])
   const script = await utils.runScript(config.hooks['build-actions'])
   if (!script) {
-    await buildActions(config, filterActions)
+    await buildActions(config, filterActions, forceBuild)
   }
   utils.runScript(config.hooks['post-app-build'])
 }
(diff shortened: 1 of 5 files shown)


shazron:
need to test for a new use case:
1. `aio app run --local`
2. use `curl` to connect to the local action, get success/error.
3. run `aio app run --local` again
4. use `curl` to connect to the local action again, Expected: success/error.

shazron:
testing the new use case: failure.

PR #559, opened 2022-07-18
Make it harder to overwrite published, Production apps

MichaelGoberling:
This pull request proposes checking the Extension Registry API to see if an App Builder application is in a published state before attempting to deploy action code or web assets. It also proposes adding a flag which will bypass this check and allow users to deploy code for published Production applications. 

## Description

These changes implement the new Console API and Extension Registry API changes introduced [here](https://github.com/adobe/aio-cli-lib-console/pull/47) so that we can determine if an App Builder application is in a published state or not. 

The `getProject` function was needed from the Console API so that we can fetch the `appId` for the project that is stored in the Extension Registry. 

A new flag `force-deploy` has been added as a part of this pull request for the `deploy` command that bypasses this check. 

## Motivation and Context

To make it more difficult to overwrite published Production App Builder applications. 

## How Has This Been Tested?

Tested locally with `npm run test` and linked version of `aio-cli-lib-console` and `aio-lib-console`. 

## Types of changes

## Checklist:


shazron:
@MichaelGoberling code conflict reported, can you resolve so the tests run?

src/commands/app/deploy.js
```diff
@@ -324,6 +342,10 @@ Deploy.flags = {
     default: true,
     exclusive: ['action']
   }),
+  'force-deploy': flags.boolean({
+    description: '[default: false] Force deploy extensions regardless of published status or workspace',
+    default: false
+  }),
   'force-publish': flags.boolean({
     description: 'Force publish extension(s) to Exchange, delete previously published extension points',
```
moritzraho, line 350: Not part of your PR but would be good to add the missing "[default:false]" in the description for consistency here too.

src/commands/app/deploy.js
```diff
@@ -324,6 +342,10 @@ Deploy.flags = {
     default: true,
     exclusive: ['action']
   }),
+  'force-deploy': flags.boolean({
+    description: '[default: false] Force deploy extensions regardless of published status or workspace',
+    default: false
+  }),
   'force-publish': flags.boolean({
```
moritzraho, line 349: Not directly part of this PR but I think there might be some confusion with `force-publish` flag unfortunately. Because you cannot publish an app via the CLI but only update the extension registry manifest in Exchange..
We should clarify this in the description, although the best would be to rename the flag.. cc @shazron @purplecabbage

src/commands/app/deploy.js
```diff
@@ -324,6 +342,10 @@ Deploy.flags = {
     default: true,
     exclusive: ['action']
   }),
+  'force-deploy': flags.boolean({
+    description: '[default: false] Force deploy extensions regardless of published status or workspace',
```
moritzraho, line 346: What about `Force deploy changes, regardless of production Workspace being published in Exchange.`

moritzraho (commented):
code lgtm, comments about flag descriptions

src/commands/app/deploy.js
MichaelGoberling: Agreed, seems like a good standard to have. Change incoming!

src/commands/app/deploy.js
MichaelGoberling: Sounds good to me 👍

src/commands/app/deploy.js
MichaelGoberling: Agreed, that flag does sound like it could use a better name. That being said, what can users tangibly accomplish by updating the extension registry manifest in Exchange? What would that be considered as an action other than `publish`?

src/commands/app/deploy.js
```diff
@@ -83,7 +84,18 @@ class Deploy extends BuildCommand {
           throw error
         }
       }
-      // 2. deploy actions and web assets for each extension
+
+      // 2. Bail if workspace is production and application status is PUBLISHED, honor force-deploy
+      if (aioConfig.project.workspace.name === 'Production' && flags.publish && !flags['force-deploy']) {
```
moritzraho, line 89: `flags.publish` in the condition here should be removed. The whole publish wording is misleading. see comment below.

src/commands/app/deploy.js
moritzraho: I think the whole publish vocabulary is misleading, what "publish" does here is just update the extension registry metadata. `force-publish` is overwriting all previous extension points config (there can be multiple extension points assigned to a workspace)

This is all confusing I realize now.

moritzraho: ok makes sense let's keep  `force-deploy` then. Renaming the flag would be a breaking change. I think we can address the code renaming and flags renaming (`force-publish` but also `(no)-publish`) as part of a separate Issue/PR.

purplecabbage:
I can't think of a better syntax for triggering this functionality so I will lets move forward with this.

shazron:
This caused a bug: #587 (am fixing)

PR #566, opened 2022-08-23
ACNA-1634 | Replace Developer Terms of Service text with help text and Developer Console url

amulyakashyap09:
## Description

Developer Terms of Service - revert the feature partially, in that when detecting that new Dev Terms are available and not accepted, we output some help text with something to this effect, and error exit:
You have not accepted the Developer Terms for your org XYZ. Please visit https://developer.adobe.com/console/home, select your XYZ org, and accept the terms.
 
I tried to have a direct url to an org, there doesn't seem to be one, just the home screen.

We delegate the acceptance to the Developer Console.

## Related Issue

[ACNA-1634](https://jira.corp.adobe.com/browse/ACNA-1634)

## Motivation and Context

The link must always be shown for legal reasons. 

## How Has This Been Tested?
- This has been tested locally
- Unit tests are passing
- No change in coverage

## Screenshots (if appropriate):
Screenshot 2022-08-23 at 1 39 27 PM

## Types of changes

## Checklist:

### DIFF
--- a/src/commands/app/init.js
+++ b/src/commands/app/init.js
@@ -18,6 +18,7 @@ const chalk = require('chalk')
 // const aioLogger = require('@adobe/aio-lib-core-logging')('@adobe/aio-cli-plugin-app:init', { provider: 'debug' })
 const { Flags } = require('@oclif/core')
 const generators = require('@adobe/generator-aio-app')
+const hyperlinker = require('hyperlinker');
 
 const { loadAndValidateConfigFile, importConfigJson } = require('../../lib/import')
 const { atLeastOne } = require('../../lib/app-helper')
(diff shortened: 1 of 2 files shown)


src/commands/app/init.js
```diff
@@ -18,6 +18,7 @@ const chalk = require('chalk')
 // const aioLogger = require('@adobe/aio-lib-core-logging')('@adobe/aio-cli-plugin-app:init', { provider: 'debug' })
 const { Flags } = require('@oclif/core')
 const generators = require('@adobe/generator-aio-app')
+const hyperlinker = require('hyperlinker')
```
shazron, line 21: since we are already including @oclif/core, why not use `CliUx.ux.url`? So we don't need another dependency.
https://github.com/oclif/core/tree/main/src/cli-ux

shazron (changes requested):

src/commands/app/init.js
amulyakashyap09: Hi @shazron ,

- oclif hyperlink functionality `CliUx.ux.url` seems to be broken, It returns promise back
- Unable to use it with template-literals .. It is not working despite many efforts
- And `oclif` has underlying same package `hyperlink` for `CliUx.ux.url` functionalilty.

Thanks,

shazron: Looks like CliUx.ux.url does not return the "hyperlinked" text, but modifies the text stream directly. Let's not waste any more time on this and use hyperlink, your original plan

shazron (approved):

PR #574, opened 2022-09-10
oclif readme --no-aliases

purplecabbage:
The new oclif README doc generator was listing the aliases as well as commands (which just clutters up the README docs), now there is a new flag --no-aliases. This is an update to the oclif package which is a dev dependency.


shazron (approved):
LGTM

PR #610, opened 2022-10-28
Call in-process hook for deploy-action with parameters

purplecabbage:
## Description

feat: Adds call to oclif hooks for `deploy-action` during run/deploy and pass… it the actions being deployed
Previous hook implementation was not fine-grained enough as when the package-script for `deploy-action` was called, there was no context of what action was being deployed and the hook-script developer had to parse the cli params themselves to see -a action-1 filters.

This hopefully allows aio-cli plugin developers a more focused hook where they implement only the required amount and not respond to every oclif hook event `init, prerun, postrun`

InProc hook is fired during the deploy-action stage of:
- `aio app deploy`
- `aio app deploy -a actionname`
- `aio app run`
- `aio app run --local`

## Types of changes

## Checklist:

### DIFF
--- a/src/commands/app/deploy.js
+++ b/src/commands/app/deploy.js
@@ -37,7 +37,7 @@ class Deploy extends BuildCommand {
 
     // if there are no extensions, then set publish to false
     flags.publish = flags.publish && !isStandaloneApp
-    let libConsoleCLI
+    let libConsoleCLI // <= this can be undefined later on, and it was not checked
     if (flags.publish) {
       // force login at beginning (if required)
       libConsoleCLI = await this.getLibConsoleCLI()
@@ -83,7 +83,7 @@ class Deploy extends BuildCommand {
       }
 
       // 2. Bail if workspace is production and application status is PUBLISHED, honor force-deploy
-      if (!isStandaloneApp && aioConfig?.project?.workspace?.name === 'Production' && !flags['force-deploy']) {
+      if (libConsoleCLI && !isStandaloneApp && aioConfig?.project?.workspace?.name === 'Production' && !flags['force-deploy']) {
         const extension = await this.getApplicationExtension(libConsoleCLI, aioConfig)
         spinner.info(chalk.dim(JSON.stringify(extension)))
         if (extension && extension.status === 'PUBLISHED') {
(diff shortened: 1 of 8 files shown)


MichaelGoberling (approved):
Like this a lot... Seems like the nextgen for app builder internal plugin hooks. Very extensible

src/commands/app/deploy.js
```diff
@@ -83,7 +83,7 @@ class Deploy extends BuildCommand {
       }
 
       // 2. Bail if workspace is production and application status is PUBLISHED, honor force-deploy
-      if (!isStandaloneApp && aioConfig?.project?.workspace?.name === 'Production' && !flags['force-deploy']) {
+      if (libConsoleCLI && !isStandaloneApp && aioConfig?.project?.workspace?.name === 'Production' && !flags['force-deploy']) {
```
shazron, line 86: `libConsoleCLI` check is not needed because of this fix: https://github.com/adobe/aio-cli-plugin-app/commit/3986d3672c5c8d8b87d2bef2ba35b6a068f1c046

src/lib/deploy-actions.js
```diff
@@ -32,6 +32,14 @@ module.exports = async (config, isLocalDev = false, log = () => {}, filter = fal
         byBuiltActions: filter
       }
     }
+    if (inprocHook) {
```
shazron, line 35: I see in the PR section `Motivation and Context`:
```
- aio app deploy [--action some-action]
  - for each action to be deployed fire an event with the config, which action is being deployed, isLocal
- aio app run [--local] 
  - for each action to be deployed
```

... that each action when deployed, the hook is fired. In my testing that is not the case, only one event is fired per deploy/run, not per action. What is the actual intent for the feature?

shazron:
Tested using this PR branch (local link) and:
```
# repo: https://github.com/shazron/oclif-hooks-test
aio plugins install shazron/oclif-hooks-test
```

src/lib/deploy-actions.js
shazron: Resolved in Slack chat: it's 1 event per deploy/run.

Issue #620, opened 2022-11-17
aio app add service does not update .env and .aio credentials


moritzraho:
**Reproduce**

- init an app with a project and workspace that has no (jwt) credentials attached.
- `aio app add service` will create the tech acct and attach service entitlements to the workspace.
- BUT the `.env` and `.aio` will not be filled with the new credentials.
- even the SERVICE_API_KEY is left empty

**Notes**
- metascopes should be updated too for workspaces that already have credentials
- this may apply to other commands that add services too (add action/extension..)

**Expected Behavior**

Should work in the same way as:
- init an app project and workspace that has (jwt) credentials attached.
- open .env

```bash
## Adobe I/O Console service account credentials (JWT) Api Key
SERVICE_API_KEY=<>
AIO_ims_contexts_aio-<>_client__id=<>
AIO_ims_contexts_aio-<>_client__secret=<>
AIO_ims_contexts_aio-<>_technical__account__email=<>@techacct.adobe.com
AIO_ims_contexts_aio-<>_technical__account__id=<>@techacct.adobe.com
AIO_ims_contexts_aio-4566206088344873777_meta__scopes=["<>"]
AIO_ims_contexts_aio-4566206088344873777_ims__org__id=<>@AdobeOrg
```

in .aio
```
[..]
"workspace": {
      "id": <>,
      "name": <>,
      "action_url": <>,
      "app_url": <>,
      "details": {
        "credentials": [
          {
            "id": <>,
            "name": <>,
            "integration_type": <>
          }
        ],
        "services": [
          {
            "code": <>
            "name": <>
          }
        ]
      },
[..]
```

**Workaround**

Run `aio app use` to sync  details from console workspace.

MichaelGoberling:
This may be a bit too into the weeds, but if we make the AddServiceCommand command extend the [Use](https://github.com/adobe/aio-cli-plugin-app/blob/master/src/commands/app/use.js) command, we can always [importConsoleConfig](https://github.com/adobe/aio-cli-plugin-app/blob/master/src/commands/app/use.js#L311) after adding the service.

aiojbot:
JIRA issue created: https://jira.corp.adobe.com/browse/ACNA-1895

PR #621, opened 2022-11-30
fix: ACNA-1895 update env as part of add service

MichaelGoberling:
## Description

Update the user's local environment with new services when running `aio app add service`. 

## Related Issue

Closes https://github.com/adobe/aio-cli-plugin-app/issues/620

## How Has This Been Tested?

Locally linked plugin with new project & `npm run test`

## Types of changes

## Checklist:


MichaelGoberling:
Tests are passing locally but failing [here in CI](https://github.com/adobe/aio-cli-plugin-app/actions/runs/3585787913/jobs/6034350339#step:7:959). Can't seem to reproduce the failures locally. This may be due to the tests being run on Ubuntu vs. Mac?

@shazron @moritzraho Thoughts?

moritzraho:
I am not sure what it's going ..  
I am not a big fan of having add services extend use, because the two commands both implement run and achieve a distinct purpose. Maybe  let `app add services` call `app use` (https://oclif.io/docs/running_programmatically#calling-commands-directly) ? (or implement a `syncWithConsole` helper used by both commands ?)

But not sure about the error, it's kinda cryptic, in seems to fail on the line that parses the flags,  maybe it's the inheritance (although it's weird that it works on mac) ?

MichaelGoberling:
PR marked as a draft while work to use a shared helper is being completed.

MichaelGoberling:
@moritzraho @shazron 

Moved the console config import functionality to `import.js` and moved all the underlying helper functions to `import-helper.js`. Now any command should be able to import the latest console config using `import.js`. 

FYI I added minimal export tests to `import.test.js` seeing as `use.test.js` does a pretty good job of testing that functionality already using the different permutations of command flags. Left a comment at the bottom of the test file indicating this.

src/lib/import-helper.js
```diff
@@ -0,0 +1,631 @@
+/*
```
moritzraho, line 1: just to be 100% sure this is just a rename from the old import ? or is it modified too ?

moritzraho (approved):
Great work ! just one double check comment, otherwise lgtm !

src/lib/import-helper.js
MichaelGoberling: It's just a rename, no real changes were made. A lot of the functions in here are only used by `import.js`, hence the naming :thumbsup:

shazron (approved):

src/lib/import-helper.js
purplecabbage: Not sure why we are just moving code around like this ... what is the benefit here?

MichaelGoberling: Yeah it is just moving code but this makes to easier to find exactly what you need to import the console config

The other file contains a lot of file processing funcs which really are just helpers

PR #650, opened 2023-04-05
feat: `aio app pack` command

shazron:
(this is based on the WIP PoC for the aio app package command)

- depends on https://github.com/adobe/aio-cli-plugin-api-mesh/pull/61 (worked around, for now)
    -  the PR has been merged, but not released in a new plugin release

## How Has This Been Tested?

- npm test

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have signed the [Adobe Open Source CLA](http://opensource.adobe.com/cla.html).
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #662, opened 2023-04-21
feat: `aio app install` command (complement to `aio app pack`)

shazron:
Complement to the `aio app pack` command in #650 
This will validate the app package created by `aio app pack`, and uncompresses the app package to your local disk.

- depends on https://github.com/adobe/aio-cli-plugin-app/pull/650 to generate the app package file to test the command

## How Has This Been Tested?

- npm test
- manual test of the command

## Types of changes

## Checklist:


shazron:
Pending completion of the schemas for `app.config.yaml` and `deploy.yaml`. Once those are done, and the tests updated for it (there are simple tests in there right now), this can be ready for review.

shazron:
`deploy.yaml` schema is complete (pending discussion with the api mesh team what properties are absolutely required).
`app.config.yaml` schema is complete as well, with minimal `runtimeManifest` checking for our purposes.

Some of the esoteric keys for `app.config.yaml` `application` key (that users will not really use or encounter, since we never document these) are defined in https://github.com/adobe/aio-cli-lib-app-config/blob/ca93bf4740342601bab1df3d1657a0b47855cadc/src/index.js#L444 (search for `singleUserConfig.`)

There needs clarification for key name validation etc in that (package name, annotation name, input name etc). For now, there is no validation for the annotations that can be set -- any annotation name can be set.

src/commands/app/install.js
```diff
@@ -0,0 +1,129 @@
...
+
+    await this.validateZipDirectoryStructure(args.path)
+    await this.unzipFile(args.path, outputPath)
+    await this.validateConfig(outputPath, 'app.config.yaml')
+    await this.validateConfig(outputPath, 'deploy.yaml')
+    await this.runTests(outputPath)
+  }
+
+  diffArray (left, right) {
+    if (!left || !right) {
```
moritzraho, line 51: Nitpick, as IIUC left, is the expected files, but if `left` is `undefined` then this returns `undefined` (meaning `left.length` will throw).
(not sure but maybe we can rename left to expected, and default it to `[]` to make it future proof)

src/commands/app/install.js
```diff
@@ -0,0 +1,129 @@
...
+    const { valid, errors } = validateJsonWithSchema(path.join(outputPath, configFileName), configFileName)
+    if (!valid) {
+      const message = `Missing or invalid keys in ${configFileName}: ${JSON.stringify(errors, null, 2)}`
+      this.error(message)
+    }
+  }
+
+  async runTests (outputPath) {
+    this.log('Running tests...')
+    const { exitCode } = await execa('aio', ['app', 'test'], { cwd: outputPath, stdio: 'inherit' })
```
moritzraho, line 101: should we call via oclif? or it doesn't matter

moritzraho (commented):
one comment and 2 nitpicks, lgtm otherwise

MichaelGoberling (commented):
This is awesome. One thing - I see that we have the full schema defined for `app.config.yaml` and we are validating those, but it seems like if the package is an extension, then we won't end up doing validation on `ext.config.yaml` (Or however the developer decides to name their extension config). Should we extend validation to inspect extension configs too?

schema/app.config.yaml.schema.json
```diff
@@ -0,0 +1,157 @@
+{
```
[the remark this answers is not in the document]
MichaelGoberling: Super cool! 

Good idea @moritzraho. We could move it to aio-cli-lib-app-config if there are other libs we can think of that would need to use it, or we could keep it here given that, in my mind at least, the next probable use case for this schema would be doing validation on `aio app deploy`

shazron:
> This is awesome. One thing - I see that we have the full schema defined for `app.config.yaml` and we are validating those, but it seems like if the package is an extension, then we won't end up doing validation on `ext.config.yaml` (Or however the developer decides to name their extension config). Should we extend validation to inspect extension configs too?

@MichaelGoberling I'm not sure if we can do this validation via the schema only since `$include` is our own definition. It will have to be an external validation where we parse `app.config.yaml` and search for any external configs to validate.
@raho one question: this `$include` only extends one level deep, or is it unbounded?

src/commands/app/install.js
shazron: Copy and paste -- I'll see if I can swap it with an oclif call

schema/app.config.yaml.schema.json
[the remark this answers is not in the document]
shazron: Good idea -- can we do this after this PR is merged? Since if not I will need to have another dependency for this PR, and testing it will be slightly inconvenient

src/commands/app/install.js
shazron: you're right, the difference should always be at least "zero" i.e the empty array []. will make the recommended changes

shazron: fixed in https://github.com/adobe/aio-cli-plugin-app/pull/662/commits/be2ca51acb754b606cec72ba70a12daacf386d24

src/commands/app/install.js
shazron: fixed with https://github.com/adobe/aio-cli-plugin-app/pull/662/commits/8f247fe7dd28c532a3d38a168222ef6e01796b95

shazron:
> This is awesome. One thing - I see that we have the full schema defined for `app.config.yaml` and we are validating those, but it seems like if the package is an extension, then we won't end up doing validation on `ext.config.yaml` (Or however the developer decides to name their extension config). Should we extend validation to inspect extension configs too?

fixed in https://github.com/adobe/aio-cli-plugin-app/pull/662/commits/724be85ff44b582ac97feb45df16f94baa23c344

schema/app.config.yaml.schema.json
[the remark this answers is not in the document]
moritzraho: sure let's just keep track of it :)

src/commands/app/install.js
moritzraho: thanks lgtm

schema/app.config.yaml.schema.json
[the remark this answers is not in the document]
shazron: https://github.com/adobe/aio-cli-plugin-app/issues/667

moritzraho:
> one question: this $include only extends one level deep, or is it unbounded?

It is unbounded...
I actually missed that, but `ext.config.yaml` is a file included in our templates, but any part of the config could be included, and `ext.config.yaml` is not a standard config file and hence shouldn't have a schema...

I think the clean way would be to coalesce all the included files into a single `app.config.yaml` during packaging... And ideally, we should have a function in `aio-cli-lib-app-config` that outputs a coalesced configuration as most of the parsing logic is there already.

shazron:
> I actually missed that, but `ext.config.yaml` is a file included in our templates, but any part of the config could be included, and `ext.config.yaml` is not a standard config file and hence shouldn't have a schema...
> 

Actually it looks like anything under `application` key in a stand-alone app (to me anyways). See the app.config.yaml schema.

> I think the clean way would be to coalesce all the included files into a single `app.config.yaml` during packaging... And ideally, we should have a function in `aio-cli-lib-app-config` that outputs a coalesced configuration as most of the parsing logic is there already.

I think we want to avoid making the packaged app any different from the existing app, just to reduce complexity -- I'm not sure how it will affect any existing config file handling that we have.

moritzraho:
Ok about modifying the existing app, but at least to validate the config we need to coalesce into one (in memory) yaml.

shazron:
> Edit: The problem is that anything could be included, even just a one liner config, so it doesn't make sense to validate ext.config.yaml because it is not a standard.

What about `runtimeManifest` though? Actually `hooks` is standard at least: https://developer.adobe.com/app-builder/docs/guides/app-hooks/

shazron:
Spoke with @raho. It's much clear now, an `$include` is just like a `C` `#include`, where the contents are replaced in, thus the included config can be anything and not have a schema. So for validation, the best way is what Moritz suggested, coalesce the config, then do the validation. I will rework the implementation for this.

shazron:
@raho @MichaelGoberling I don't think I should do the validation of app.config.yaml with $includes here in the app plugin:
1. firstly, I can't get access to the "pure" config since I can only load everything via this one export: https://github.com/adobe/aio-cli-lib-app-config/blob/ca93bf4740342601bab1df3d1657a0b47855cadc/src/index.js#L607
2. As discussed in a thread above, this validation should be in the @adobe/aio-cli-lib-app-config anyway, when loading the configs, and tracked in https://github.com/adobe/aio-cli-plugin-app/issues/667 and https://github.com/adobe/aio-cli-lib-app-config/issues/11

This could be a fast follow issue to implement, and #667 will include using loadConfig() (with an option to specify where the config is, or I can just change the current working directory) -- either that or have multiple exports (and publishing a major version update). 

If there are no other issues (besides punting this off to #667) I believe all issues are resolved.

moritzraho:
@shazron sounds good!
However, if possible, I would remove the `ext.config.yaml` schema

shazron:
Oh yeah I missed this one , will do
________________________________

shazron:
> @shazron sounds good! However, if possible, I would remove the `ext.config.yaml` schema

removed in https://github.com/adobe/aio-cli-plugin-app/pull/662/commits/1b0ab5836e66db85bebbaf7b31ad95ad9d86a422

shazron:
I'll merge this first thing tomorrow and do more testing, which will include #650

PR #664, opened 2023-04-26
In proc hooks

purplecabbage:
## Description

Before calling a package/project script, we will try to require it in.
Our default implementation has been similar to `npm run <script-name>`, which works fine, but is not passed any context.
Before calling the default implementation of `runScript`, we call `runInProcess` passing the loaded appConfig.

`runInProcess` will attempt to require the string it is passed

ex.
`await runInProcess('hooks/pre-app-build', config)`

'hooks/pre-app-build' is:
1. expected to be a javascript file in the project
2. it should export an async function (which will receive a config object)

If both conditions are met the `runInProcess` will call the function with the config and return whatever the function returns.

If require fails, `runInProcess` will fall back to the default behavior and call 'runScript'

## Motivation and Context

Some extensibility use-cases are better handled with a few scripts that live in the actual project instead of trying to make the `aio app` command handle every possible future feature.

## Types of changes

## Checklist:


src/lib/app-helper.js
```diff
@@ -66,7 +66,12 @@ async function installPackages (dir, options = { spinner: null, verbose: false }
...
 }
 
-/** @private */
+// Is this still used? it is exported, but there are no references to it -jm
+/**
+ * @param {string} scriptName  npm script name
+ * @param {string} dir directory to run npm script in
+ * @param {string[]} cmdArgs args to pass to npm script
+ */
 async function runPackageScript (scriptName, dir, cmdArgs = []) {
```
shazron, line 75: if it's exported, usually i would say we can't remove it without bumping a major, but considering it's an app helper script (internal)I think it would be fine to remove.

src/lib/build-actions.js
```diff
@@ -21,10 +21,10 @@ const { buildActions } = require('@adobe/aio-lib-runtime')
  * @param {boolean} [forceBuild=false] force a build (skip file changed hash check)
  */
 module.exports = async (config, filterActions, forceBuild = false) => {
-  utils.runScript(config.hooks['pre-app-build'])
-  const script = await utils.runScript(config.hooks['build-actions'])
+  runInProcess(config.hooks['pre-app-build'], config)
```
shazron, line 24: do we need to await here, if not there is a possibility that `build-actions` is run before `pre-app-build` runs? Similarly for `post-app-build`

src/lib/deploy-actions.js
```diff
@@ -23,8 +23,8 @@ const { deployActions } = require('@adobe/aio-lib-runtime')
  */
 /** @private */
 module.exports = async (config, isLocalDev = false, log = () => {}, filter = false, inprocHook) => {
-  utils.runScript(config.hooks['pre-app-deploy'])
-  const script = await utils.runScript(config.hooks['deploy-actions'])
+  runInProcess(config.hooks['pre-app-deploy'], config)
```
shazron, line 26: similar comment like I had regarding `pre-app-build` above (for pre-app-deploy and post-app-deploy)

shazron (changes requested):

src/lib/deploy-actions.js
purplecabbage: Done

src/lib/build-actions.js
purplecabbage: Fixed

Issue #699, opened 2023-07-25
aio app pack: Fails if api-mesh plugin present but no api mesh on the project


MichaelGoberling:
**Describe the bug**
The `aio app pack` command fails if you have the api-mesh plugin installed but there is no api mesh configured for the project you are packing from

**To Reproduce**
1. Go to your App Builder project workspace
2. Make sure api-mesh plugin is installed `aio plugins install api-mesh`
3. Try to pack with `aio app pack` 
4. See the following error 
```
➜  color-web-app-excshell git:(main) ✗ aio app pack 
Pre-release warning: This command is in pre-release, and not suitable for production.
✔ Created package artifacts folder 'app-package'
✔ Copied project files
✖ Command failed with exit code 2: aio api-mesh get
 ›   Error: Unable to get mesh config. No mesh found for Org(53444) -> 
 ›   Project(4566206088344868334) -> Workspace(4566206088344873890). Please 
 ›   check the details and try again.
Selected organization: Adobe IO
Selected project: mgtestutils
Select workspace: Stage
 ›   Error: Command failed with exit code 2: aio api-mesh get
 ›    ›   Error: Unable to get mesh config. No mesh found for Org(53444) -> 
 ›    ›   Project(4566206088344868334) -> Workspace(4566206088344873890). Please 
 ›    ›   check the details and try again.
 ›   Selected organization: Adobe IO
 ›   Selected project: mgtestutils
 ›   Select workspace: Stage
```

**Expected behavior**
1. Go to your App Builder project 
2. Make sure api-mesh plugin is installed `aio plugins install api-mesh`
3. Try to pack with `aio app pack` 
4. Packing succeeds, the call to api-mesh does not fail the pack command

**Screenshots**

**Desktop (please complete the following information):**
 - aio cli 9.3.0

PR #700, opened 2023-07-25
fix: handle mesh config 404 error

MichaelGoberling:
## Description

If the user has the api-mesh plugin installed and runs `aio app pack` on a project without an api-mesh config, packing will fail with a not found error. This PR proposes ignoring these errors and continuing to pack

## Related Issue

Closes https://github.com/adobe/aio-cli-plugin-app/issues/699

## Motivation and Context

So `aio app pack` works with all plugins installed

## How Has This Been Tested?

Locally linked plugin, `npm run test`

## Types of changes

## Checklist:

### DIFF
--- a/src/commands/app/pack.js
+++ b/src/commands/app/pack.js
@@ -135,13 +135,23 @@ class Pack extends BaseCommand {
     // TODO: send a PR to their plugin to have a `--json` flag
     const command = await this.config.findCommand('api-mesh:get')
     if (command) {
-      this.spinner.start('Getting api-mesh config...')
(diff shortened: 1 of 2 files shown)


MichaelGoberling:
@shazron @purplecabbage @moritzraho 

Opening this up as a draft for now, because I'm wondering if there is a case where we should fail if the api-mesh is not found. For instance, can the user define the api-mesh config in app.config.yaml or in another local file and then when we go to fetch the api-mesh config that means there is a mismatch between the server and local config?

moritzraho:
For now, apimesh is not exposed in `app.config.yaml` (would be good to add it eventually)

moritzraho (approved):

PR #710, opened 2023-08-09
Update aio app add events to fetch events templates from Template Registry

sangeetha5491:
## Types of changes

## Checklist:


src/commands/app/add/event.js
```diff
@@ -29,27 +31,63 @@ class AddEventCommand extends AddCommand {
...
-    const eventsGen = env.instantiate(generators['add-events'], {
-      options: {
-        'skip-prompt': flags.yes,
-        'action-folder': actionFolder,
-        'config-path': configData.file,
-        'full-key-to-manifest': configData.key,
-        // force overwrites, no useless prompts, this is a feature exposed by yeoman itself
-        force: true
+    const templateOptions = {
+      'skip-prompt': false,
```
moritzraho, line 41: why not flags.yes?

src/commands/app/add/event.js
```diff
@@ -29,27 +31,63 @@ class AddEventCommand extends AddCommand {
     if (entries.length > 1) {
       this.error('Please use the \'-e\' flag to specify to which implementation you want to add events to.')
     }
+
     const configName = entries[0][0]
     const config = entries[0][1]
     const actionFolder = path.relative(config.root, config.actions.src)
     const configData = this.getRuntimeManifestConfigFile(configName)
```
moritzraho, line 38: nitpick: should we rename`configData` => `runtimeData` (to contrasts with eventsData below)

src/commands/app/add/event.js
```diff
@@ -29,27 +31,63 @@ class AddEventCommand extends AddCommand {
...
-        'action-folder': actionFolder,
-        'config-path': configData.file,
-        'full-key-to-manifest': configData.key,
-        // force overwrites, no useless prompts, this is a feature exposed by yeoman itself
-        force: true
+    const templateOptions = {
+      'skip-prompt': false,
+      'action-folder': actionFolder,
+      'config-path': configData.file,
+      'full-key-to-manifest': configData.key
```
moritzraho, line 44: if the events template API is still open for changes I would use `full-key-to-runtime-manifest` in the new template. Otherwise nvm

src/BaseCommand.js
```diff
@@ -177,7 +193,11 @@ class BaseCommand extends Command {
 
 BaseCommand.flags = {
   verbose: Flags.boolean({ char: 'v', description: 'Verbose output' }),
-  version: Flags.boolean({ description: 'Show version' })
+  version: Flags.boolean({ description: 'Show version' }),
+  allowEventsTemplates: Flags.boolean({
```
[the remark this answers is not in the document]
moritzraho: oh and 
3. if we use a flag we should use dashes instead of camelCase `allow-events-templates`

src/commands/app/add/event.js
sangeetha5491: I let it be consistent with other templates where full-key-to-manifest meant runtimeManifest.

src/BaseCommand.js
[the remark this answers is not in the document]
sangeetha5491: @purplecabbage any thoughts on the approach to use here?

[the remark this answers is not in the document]
sangeetha5491: `allow-events-templates` i can use this if we decide to go with the flag approach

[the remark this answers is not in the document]
sangeetha5491: if its an env variable, how and where do i set it?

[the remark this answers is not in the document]
sangeetha5491: made allow-events-templates flag local to the add event template command

src/commands/app/add/event.js
sangeetha5491: for events templates, we always need the skip-prompt to be false, true will not make sense as there are no default values that can be added to the app.

src/commands/app/add/event.js
sangeetha5491: also, the ActionGenerator which is extended by the EventsGenerator expects the full-key-to-manifest

src/commands/app/add/event.js
sangeetha5491: does not matter since both a local to the respective methods.

sangeetha5491: done

src/commands/app/add/event.js
moritzraho: ok then maybe let's add a note to the flag that it will not have any effect when allow-events-template is set

src/commands/app/add/event.js
moritzraho: thanks

src/commands/app/add/event.js
moritzraho: ok then nvm 👍

moritzraho (approved):

src/BaseCommand.js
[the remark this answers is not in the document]
shazron: Definitely all flags should be kebab-case, thanks.

If the flag is temporary, we should never add it as a flag because it will break people in the future unless we bump a major version, and add a deprecation period - so we should try to avoid this as much as possible.

We haven't done this before but we could prefix `experimental-` to the flag name, much like how node does it for their experimental flags, which have no guarantees. Either using this prefix or the env var will break users once it is removed, but at least with the flag it will be more explicit.

Setting the flag to automatically pick up an env var is part of the oclif flag definition, see https://oclif.io/docs/flags (`env` key)

WDYT?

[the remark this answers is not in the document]
sangeetha5491: sure, ill add the experimental- prefix to the flag.

src/commands/app/add/event.js
sangeetha5491: thanks, added. also added a related issue to be fixed in https://github.com/adobe/generator-app-common-lib/issues/38

moritzraho: nice thanks

PR #734, opened 2023-10-05
fix: ACNA-2561 - app pack command should exclude hidden files

shazron:
Currently the files to be packed are defined by `npm pack --dry-run --json` and exclusions are defined by the `.gitignore` file. However, if the app folder does not have `.gitignore`, certain sensitive files may still be included. This PR will exclude these files no matter what.

These files are excluded from being packed:
1. any "hidden" file by convention (file that starts with a '.')
2. any junk file as defined by https://www.npmjs.com/package/junk

## How Has This Been Tested?

npm test

## Types of changes

## Checklist:

### DIFF
--- a/src/commands/app/pack.js
+++ b/src/commands/app/pack.js
@@ -21,6 +21,7 @@ const { loadConfigFile, writeFile } = require('../../lib/import-helper')
 const { getObjectValue } = require('../../lib/app-helper')
 const ora = require('ora')
 const chalk = require('chalk')
+const junk = require('junk')
 
 // eslint-disable-next-line node/no-missing-require
 const libConfigNext = require('@adobe/aio-cli-lib-app-config-next')
@@ -250,6 +251,8 @@ class Pack extends BaseCommand {
     return files
       .map(file => file.path)
       .filter(file => !filesToExclude.includes(file))
+      .filter(junk.not) // no junk files like .DS_Store
+      .filter((file) => !/^\..*/.test(file)) // no files that start with a '.'
   }
 
   /**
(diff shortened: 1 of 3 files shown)


sandeep-paliwal (approved):

src/commands/app/pack.js
```diff
@@ -250,6 +251,8 @@ class Pack extends BaseCommand {
     return files
       .map(file => file.path)
       .filter(file => !filesToExclude.includes(file))
+      .filter(junk.not) // no junk files like .DS_Store
+      .filter((file) => !/^\..*/.test(file)) // no files that start with a '.'
```
moritzraho, line 255: should we ignore gitignore files instead.. or is this a quick fix?

moritzraho (approved):

src/commands/app/pack.js
moritzraho, line 255: Should we stdout/document this behavior somewhere?

shazron: good idea - maybe as a debug log. Maybe stdout for files we *include* would be better instead? (like when running npm pack in a npm package folder)

purplecabbage (approved):

src/commands/app/pack.js
shazron: added in https://github.com/adobe/aio-cli-plugin-app/pull/734/commits/834b482413b787b1ae79860fe8012c9614e64996:
1. debug logging of omitted hidden dotfiles
2. debug logging of omitted junk files
3. stdout logging of files that will be packed

PR #736, opened 2023-10-11
fix: ACNA-2562 - artifacts and zip output should be in a .gitignored dist folder

shazron:
These files have been moved:
1. `app-package` artifacts folder has been moved to `dist/app-package`
2. the default output `app.zip` has been moved to `dist/app.zip`.

This has been added:
1. all extension `dist` folders are excluded from the pack
2. the `dist` folder itself is excluded from the pack (just in case the .gitignore value is not there)
3. the `dist/app-package` folder will be deleted after the pack

## How Has This Been Tested?

1. npm test
2. `aio app pack` on an App Builder project
## Types of changes

## Checklist:

### DIFF
--- a/src/commands/app/pack.js
+++ b/src/commands/app/pack.js
@@ -26,10 +26,11 @@ const junk = require('junk')
 // eslint-disable-next-line node/no-missing-require
 const libConfigNext = require('@adobe/aio-cli-lib-app-config-next')
 
+const DIST_FOLDER = 'dist'
 const DEFAULTS = {
-  OUTPUT_ZIP_FILE: 'app.zip',
-  ARTIFACTS_FOLDER: 'app-package',
-  DEPLOY_YAML_FILE: 'deploy.yaml'
+  OUTPUT_ZIP_FILE_PATH: path.join(DIST_FOLDER, 'app.zip'),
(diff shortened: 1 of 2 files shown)


arjuncooliitr (approved):

shazron:
FYI @sangeetha5491 - slight change to the artifacts folder being passed in to the hook.

MichaelGoberling (approved):
Changes LGTM!

Follow up Q: Is there a known use case for being able to configure the output of the app zip? It seems useful, but I was wondering if we could maybe reduce the chance a dev accidentally packages `foo/app.zip` if the app zip _always_ goes in `dist/app.zip`

shazron:
> Changes LGTM!
> 
> Follow up Q: Is there a known use case for being able to configure the output of the app zip? It seems useful, but I was wondering if we could maybe reduce the chance a dev accidentally packages `foo/app.zip` if the app zip _always_ goes in `dist/app.zip`

The use case is if an ISV wants to automate and put all packages into a known folder, and they can name it immediately to what they want, say for archiving. Say `--output ~/Desktop/my-app-packages/my-project-abc-2023-10-12.zip`

PR #739, opened 2023-10-20
Enable deployment of blackboxes to local openwhisk via `docker` manifest property

hparra:
## Description

- When deploying actions, `aio` checks whether kind is supported by cli, and if not double-checks api [1].
- When deploying locally using `aio run app --local`,  openwhisk uses _runtimes.json_ to configure which kinds are supported.
- Even though blackboxes are not restricted locally, `aio` prevents deployment because blackbox is not returned in supported kinds.
- We fix this by adding blackbox kind to runtimes.json for local openwhisk.

[1] https://github.com/adobe/aio-lib-runtime/blob/e8815561e7d7e06e75e8f4d8cc6ef69da12fb7d2/src/utils.js#L1494

## Related Issue

#740

## Motivation and Context

I was developing a blackbox image locally before I had access to feature. Each time the actions are redeployed on code change I also have to rerun `wsk action update ... --docker ...` to update the container, but this seams unnecessary as the manifest already supports this.

## How Has This Been Tested?

This has been tested locally by monkeypatching these changes in local node_modules and successfully deploying a blackbox to my local openwhisk using the `docker` manifest property only.

## Types of changes

## Checklist:

### DIFF
--- a/bin/openwhisk-standalone-config/runtimes.json
+++ b/bin/openwhisk-standalone-config/runtimes.json
@@ -2,6 +2,12 @@
   "runtimes": {
     "nodejs": [
       {
+        "kind": "blackbox",
+        "image": {
+          "name": "blackbox"
+        }
+      },
+      {
         "kind": "nodejs:10",
         "image": {
           "prefix": "adobeapiplatform",


bin/openwhisk-standalone-config/runtimes.json
```diff
@@ -1,6 +1,12 @@
 {
   "runtimes": {
     "nodejs": [
+      {
+        "kind": "blackbox",
+        "image": {
+          "name": "blackbox"
+        }
+      },
```
shazron, line 9: Thanks! I think it's best to update this script to add this entry statically, since that's how we generate this json: https://github.com/adobe/aio-cli-plugin-app/blob/master/bin/openwhisk-standalone-config/get-runtimes.js

related: https://github.com/adobe/aio-cli-plugin-app/pull/741

shazron (changes requested):

hparra:
@shazron Agreed but now taking different approach. Change in https://github.com/adobe/aio-lib-runtime/pull/168 fixes both local and cloud. I have updated https://github.com/adobe/aio-cli-plugin-app/issues/740. I will reopen this if necessary.

PR #811, opened 2024-08-22
Optimize build and deploy

purplecabbage:
## Description

This changes the default for flag --force-build to false
RuntimeLib will include additional checks to not build actions it has already built, and not deploy actions already deployed.  In order for users to see these benefits, the defaults are to use the optimized workflow and reduce time to build/deploy/run

If an action has changed, it will be built during app build/deploy/run
If an action has changed ( or action inputs including from $env or action descriptors ) it will be deployed during app deploy/run

## Types of changes

## Checklist:

### DIFF
--- a/src/commands/app/build.js
+++ b/src/commands/app/build.js
@@ -70,14 +70,16 @@ class Build extends BaseCommand {
     }
 
     if (flags.actions) {
-      if (config.app.hasBackend && (flags['force-build'] || !fs.existsSync(config.actions.dist))) {
+      // removed flags['force-build'] || as it is always forced at this point, we need to check to decide what to build
+      // if no backend, we skip the build
+      if (config.app.hasBackend) { //  && (flags['force-build'] || !fs.existsSync(config.actions.dist))) {
         try {
           let builtList = []
           const script = await runInProcess(config.hooks['build-actions'], config)
           aioLogger.debug(`run hook for 'build-actions' for actions in '${name}' returned ${script}`)
           spinner.start(`Building actions for '${name}'`)
           if (!script) {
-            builtList = await RuntimeLib.buildActions(config, filterActions, true)
+            builtList = await RuntimeLib.buildActions(config, filterActions, flags['force-build']) // skipCheck is false
           }
           if (builtList.length > 0) {
             spinner.succeed(chalk.green(`Built ${builtList.length} action(s) for '${name}'`))
@@ -85,7 +87,7 @@ class Build extends BaseCommand {
             if (script) {
               spinner.fail(chalk.green(`build-action skipped by hook '${name}'`))
             } else {
-              spinner.fail(chalk.green(`No actions built for '${name}'`))
+              spinner.info(chalk.green(`No actions built for '${name}'`))
             }
           }
           aioLogger.debug(`RuntimeLib.buildActions returned ${builtList}`)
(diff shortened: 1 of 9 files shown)


src/commands/app/build.js
```diff
@@ -70,22 +70,24 @@ class Build extends BaseCommand {
     }
 
     if (flags.actions) {
-      if (config.app.hasBackend && (flags['force-build'] || !fs.existsSync(config.actions.dist))) {
+      // removed flags['force-build'] || as it is always forced at this point, we need to check to decide what to build
+      // if no backend, we skip the build
+      if (config.app.hasBackend) { //  && (flags['force-build'] || !fs.existsSync(config.actions.dist))) {
```
moritzraho, line 75: q: I know this was here before but `flags.actions && !config.app.hasBackend` should just exit right?

moritzraho (approved):

src/commands/app/build.js
purplecabbage: yeah, I need to clean this up, if there is no backend then there is nothing to build so we simply output 'no backend or a build already exists' but it should probably just be 'nothing to build, no backend'

src/lib/run-dev.js
```diff
@@ -82,7 +82,7 @@ async function runDev (config, dataDir, options = {}, log = () => {}, inprocHook
 
       // build and deploy actions
       log('building actions..')
-      await buildActions(devConfig, null, true /* force build */)
+      await buildActions(devConfig, null, false /* force build */)
```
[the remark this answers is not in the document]
purplecabbage: We don't force build by default because we are tracking changes.  Force build would build regardless of whether things have changed or not.

src/commands/app/build.js
purplecabbage: cleaned up!

src/lib/actions-watcher.js
```diff
@@ -91,6 +91,8 @@ function createChangeHandler (watcherOptions) {
     try {
       aioLogger.debug(`${filePath} has changed. Redeploying actions.`)
       const filterActions = getActionNameFromPath(filePath, watcherOptions)
+      // this is happening, but its not showing up because verbose is usually off
+      // this might be more important and worthy of signalling to the user
```
shazron, line 95: I think this is important if we can define the importance of the "non-action file" is (more files are covered by your new changes) - the user might not know why their changes are not showing up otherwise

src/lib/run-dev.js
```diff
@@ -152,10 +151,16 @@ async function runDev (config, dataDir, options = {}, log = () => {}, inprocHook
       devConfig.app.hasFrontend = false
     }
 
-    log('setting up vscode debug configuration files...')
-    const vscodeConfig = vscode(devConfig)
-    await vscodeConfig.update({ frontEndUrl })
-    cleanup.add(() => vscodeConfig.cleanup(), 'cleaning up vscode debug configuration files...')
+    // todo: remove vscode config swapping, dev command uses a persistent file so we don't need this.
```
shazron, line 154: we should remove --local with a fast follow PR and remove this in that PR

src/commands/app/build.js
```diff
@@ -163,8 +166,8 @@ Build.flags = {
     allowNo: true
   }),
   'force-build': Flags.boolean({
```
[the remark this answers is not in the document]
purplecabbage: The default behavior of buildActions has changed, it will always build actions that it determines 'need' to be built.  In this case, force-build can still be used if an undetectable change has been made.  For example, if an action imports|requires a module and the root application's package.json dependency version was updated _(and `npm i` was run)_ this change would not be noticed and the actions would not be rebuilt.
Effectively `force-build` could be thought of as clean-and-build

shazron (approved):

PR #814, opened 2024-10-08
Add app/version useragent for tracking clients in runtime

purplecabbage:
## Types of changes

## Checklist:

### DIFF
--- a/src/BaseCommand.js
+++ b/src/BaseCommand.js
@@ -42,7 +42,13 @@ class BaseCommand extends Command {
     await super.init()
     // setup a prompt that outputs to stderr
     this.prompt = inquirer.createPromptModule({ output: process.stderr })
-    process.env.__OW_USER_AGENT = 'aio-cli-plugin-app@' + require('../package.json').version
+
+    // set User-Agent for runtime calls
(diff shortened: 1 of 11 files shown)


src/BaseCommand.js
```diff
@@ -42,6 +42,7 @@ class BaseCommand extends Command {
     await super.init()
     // setup a prompt that outputs to stderr
     this.prompt = inquirer.createPromptModule({ output: process.stderr })
+    process.env.__OW_USER_AGENT = 'aio-cli-plugin-app@' + require('../package.json').version
```
shazron, line 45: just a reviewer note: even though it's not documented in oclif, the command's `init` function is always called before a `run` is called: https://github.com/oclif/core/blob/68110ee2a4839f0b3ab299ae775d3af1f0c107bc/src/command.ts#L359

If we want to add more details like a browser user agent, we can get it from `.config.versionDetails()` (but this is not documented): https://github.com/oclif/core/blob/68110ee2a4839f0b3ab299ae775d3af1f0c107bc/src/config/config.ts#L187

shazron (approved):

MichaelGoberling (approved):

src/BaseCommand.js
purplecabbage: Done! Thanks

PR #853, opened 2025-02-28
fix: deploy - output the error from the audit log api if `--verbose` is set

shazron:
## How Has This Been Tested?

- local link the plugin
- try to deploy with `--verbose`, it should error in the audit log api (there is a bug/service issue currently), and show the exception message

## Types of changes

## Checklist:

### DIFF
--- a/src/commands/app/deploy.js
+++ b/src/commands/app/deploy.js
@@ -112,6 +112,9 @@ class Deploy extends BuildCommand {
               await sendAuditLogs(cliDetails.accessToken, assetDeployedLogEvent, cliDetails.env)
             } catch (error) {
               this.warn('Error: Audit Log Service Error: Failed to send audit log event for deployment.')
+              if (flags.verbose) {
+                this.warn(error.message)
+              }
             }
           }
         }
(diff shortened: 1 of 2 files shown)


pru55e11 (approved):

src/commands/app/deploy.js
```diff
@@ -112,6 +112,9 @@ class Deploy extends BuildCommand {
               await sendAuditLogs(cliDetails.accessToken, assetDeployedLogEvent, cliDetails.env)
             } catch (error) {
               this.warn('Error: Audit Log Service Error: Failed to send audit log event for deployment.')
```
purplecabbage, line 114: I kinda think nothing should be output if verbose is false ... there is no reason the developer should know or care if their action was or was not tracked, the knowledge is of no benefit to them, and the fact that it is a warning could cause undue alarm.

purplecabbage (approved):

src/commands/app/deploy.js
pru55e11: I agree with @purplecabbage - The error is not actionable by the user and we shouldn't surface it except in debug

shazron: will make the change

shazron: fixed in [3115099](https://github.com/adobe/aio-cli-plugin-app/pull/853/commits/3115099aaf35641286df2fb6bbbda88def390579)

